PR #44, opened 2020-05-17
Feature: TTS word by word highlight with Splitter

aferditamuriqi:


src/modules/TTS/TTSModule.ts
```diff
@@ -0,0 +1,291 @@
...
+        console.log(voices)
+
+        // use publication language
+        if ((self.annotationModule.delegate.tts.voice && self.annotationModule.delegate.tts.voice.usePublication) || self.annotationModule.delegate.tts.voice == undefined) {
+            utterance.voice = voices.filter((el: any) => el.lang.startsWith(self.annotationModule.delegate.publication.metadata.language[0]) || el.lang.endsWith(self.annotationModule.delegate.publication.metadata.language[0].toUpperCase()))[0]
+        }
+        // if no voice, then use configured language
+        if (self.annotationModule.delegate.tts.voice && (utterance.voice == undefined || utterance.voice == null)) {
+            utterance.voice = voices.filter((el: any) => el.lang == self.annotationModule.delegate.tts.voice.lang && el.name == self.annotationModule.delegate.tts.voice.name)[0]
+        }
```
bgoldowsky: This is not a big deal for us right now, but for future-proofing perhaps the "preferred voice" preference should be a map of languages to voices rather than a single voice.  So I could have a favorite voice for English, a different one for French, etc.

src/modules/TTS/TTSModule.ts
```diff
@@ -0,0 +1,291 @@
...
+                                        }
+                                        splittingWord.style.background = self.tts.color    
+
+                                    }
+
+                                    if (!verticalScroll && self.annotationModule.delegate.tts.autoScroll) {
+                                        splittingWord.scrollIntoView({
+                                            block: "center",
+                                            behavior: "smooth",
+                                        })
```
bgoldowsky: Just curious, I see it will auto scroll, will it also move across pages in paged mode?

src/modules/TTS/TTSModule.ts
```diff
@@ -0,0 +1,291 @@
...
+                }
+            }
+        }
+            
+        utterance.onend = function () {      
+            if (IS_DEV) console.log("utterance ended");
+            self.annotationModule.highlighter.doneSpeaking(true)
+            if (self.annotationModule.delegate.tts.enableSplitter) {
+                var splittingWord = splittingResult[splittingResult.length-1] as HTMLElement
+                splittingWord.style.removeProperty("background")
```
bgoldowsky: Per today's discussion, perhaps it could remember the word where it stopped so that a 'resume' was possible. Also this could enable switching rate or pitch on the fly, by stopping and then resuming speech at the place where it was stopped.

src/modules/TTS/TTSModule.ts
```diff
@@ -0,0 +1,291 @@
...
+                                            }
+                                        });
+                                        if (elements.length > 1) {                                            
+                                            whitespaces1.forEach(element => {
+                                                element.style.background = self.tts.color
+                                            });
+                                        }
+
+
+                                    } else if (self.annotationModule.delegate.tts.highlight == "word") {
```
bgoldowsky: Option names should be consistent - lines/words or line/word.

bgoldowsky:
It would be nice to have settings for both **foreground** text color and **background** color for the word/line that's highlighted during TTS.

src/modules/TTS/TTSModule.ts
aferditamuriqi: this is actually from the splitting library, but i have to fix another issue with that library

aferditamuriqi: actually never mind, it's my mistake, it should be words

src/modules/TTS/TTSModule.ts
aferditamuriqi: In paged mode it won't move right now across pages. 
that reminds me though that it should move to the next page if the ned is reached.

src/modules/TTS/TTSModule.ts
aferditamuriqi: ok, noted, yes we talked about this.

aferditamuriqi:
> It would be nice to have settings for both **foreground** text color and **background** color for the word/line that's highlighted during TTS.

that's a good idea, i can add that.

src/modules/TTS/TTSModule.ts
aferditamuriqi: moved to a task in AC

src/modules/TTS/TTSModule.ts
aferditamuriqi: In progress

PR #46, opened 2020-05-25
Initial settings and several api updates 

aferditamuriqi:


src/navigator/IFrameNavigator.ts
```diff
@@ -684,10 +684,13 @@ export default class IFrameNavigator implements Navigator {
             let startUrl: string | null = null;
             if (startLink && startLink.href) {
                 startUrl = this.publication.getAbsoluteHref(startLink.href);
-
             }
 
             if (lastReadingPosition) {
+                const linkHref = this.publication.getAbsoluteHref(lastReadingPosition.href);
+                console.log(lastReadingPosition.href)
```
swandog30: Do you want an if (IS_DEV) on these two console.log statements?

aferditamuriqi: done.

PR #49, opened 2020-05-26
Selection-level highlighting of TTS words (initial, for comment)

waharnum:
This adds by-word highlighting for selections that fill within the same block (selections spanning multiple blocks are more complex and will need further exploration). It's rough right now, but at a point where I think comments would be useful. In particular, I am fairly new to TypeScript, and want to make sure code I contribute is appropriately idiomatic in how it's being used.

Thoughts/observations:
- the raw `Range` was added to the `ISelectionInfo` interface. I suspect this may not be appropriate long-term, and the `ISelectionInfo` / `IRangeInfo` interfaces should have additional functionality added to support what's done with the `Range` in the `speak` function of `TTSModule`.
- the approach taken uses `Splitting.js` in a somewhat different way, and does some DOM manipulation to temporarily replace the text within the selection with word-splitting markup, then restore the original content.
- I think initial refactoring should focus on extracting a common model of operation for the use case (highlighting individual words as TTS speaks); this is complicated by the realities of how the Selection/Range features work, but for the moment there appear to be three main areas to consider:
  - speaking and highlighting the entire current resource, word by word (implemented)
  - speaking and highlighting a selection within a single block (implemented)
  - speaking and highlighting a selection spanning multiple blocks (not implemented)

### DIFF
--- a/src/model/Locator.ts
+++ b/src/model/Locator.ts
@@ -58,6 +58,7 @@ export interface ISelectionInfo {
     cleanText: string;
     rawText: string;
     color: string;
+    range: Range;
 }
 
 export interface IRangeInfo {
--- a/src/modules/TTSModule.ts
+++ b/src/modules/TTSModule.ts
@@ -28,19 +28,81 @@ export default class TTSModule implements ReaderModule {
     }
     cancel() {
         this.synth.cancel()
-    }
+    }        
 
-    speak(selectionInfo: ISelectionInfo | undefined ): any {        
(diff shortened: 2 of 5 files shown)


aferditamuriqi:
@waharnum I am merging this into the TTS branch, for now until the Splitter POC is done will leave it in that branch and we'll make decisions after the Splitter POC.

aferditamuriqi (approved):

PR #58, opened 2020-06-23
Feature/highlight var passthrough

mbrambilla:
@aferditamuriqi 

Spent a bit of time looking at if it might be possible to pass a CSS variable through R2D2BC.

There are a few areas that are not completed yet, such as the underline highlight variant since that involves a compound color scheme, but this might be useful as a starting point.

- Concept was tested by passing a CSS variable through on `initialAnnotationColor`.
- Used a `.hover` class to allow hover state of highlight to be seen upstream application
- Reference against: https://github.com/cast-org/clusive/tree/feature/CSL-292-highlight-themes


aferditamuriqi (approved):
looks good to me. will need to merge and test with some of the work that i am doing right now.

aferditamuriqi:
Looks good, i'll be merging this PR, and add the an example css and usage into the sample html.

PR #61, opened 2020-06-29
Added example css style as injectable for annotation color.

aferditamuriqi:
(this will need more expansion for all other colors, but for the initial color it works fine.)


swandog30 (approved):
Looks good.  .75 alpha on the initial highlight color is a good idea, full on yellow can be intense

PR #101, opened 2020-11-24
Improvements: callbacks, apis and tts fixes

aferditamuriqi:
- fixes scroll issue on iOS Safari, prevent resize trigger if scrolling is in motion
- fixes default language on iOS Safari
- added callback for: resourceReady, selectionMenuOpen, selectionMenuClose
- added api for: applyAttributes
- several html and css cleanups

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@d-i-t-a/reader",
-  "version": "1.5.0",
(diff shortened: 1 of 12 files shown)


bgoldowsky:
From what I can see, looks good. Will test out when I have a chance.

bgoldowsky (dismissed):

PR #149, opened 2021-03-22
TS declarations

kristojorg:
This PR is to include Typescript declarations in the build output. To do this I enabled `declarations` in the `tsconfig.json`, and then had to update where the `package.json` was telling consumers the main `.js` and `.d.ts` files were. As a consequence, you can now simply import from the package root:
```js
import D2Reader from '@d-i-t-a/reader';
```
instead of importing from the `dist` folder directly:
```js
import D2Reader from '@d-i-t-a/reader/dist/reader';
```
and you now get full type defs when using the library:
image

I think this should be fully backwards compatible because you can still import from `@d-i-t-a/reader/dist/reader`, but note that typescript will not find the declaration files in that case, because they are under the name `index.d.ts`. I couldn't find a setting to rename them to `reader.d.ts`, but if you import from the package root then the `"types": "dist/index.d.ts"` property in `package.json` fixes that issue by associating the two files. 

One way to fix that would be to change the `output.filename` in webpack to `index.js` for the reader, then you would automatically find the `index.d.ts` no matter how it was imported, but that would be a breaking change so I avoided it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,8 +4,8 @@
   "description": "A viewer application for EPUB files.",
   "repository": "https://github.com/d-i-t-a/R2D2BC",
   "license": "Apache-2.0",
-  "main": "src/index.js",
-  "typings": "src/index.ts",
+  "main": "dist/reader.js",
+  "typings": "dist/index.d.ts",
   "engines": {
     "node": ">8.0.0"
   },
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -32,6 +32,7 @@
     "moduleResolution": "node",
     "outDir": "./dist/",
     "sourceMap": true,
+    "declaration": true,
     "preserveConstEnums": true,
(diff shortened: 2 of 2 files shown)


aferditamuriqi:
Hi @kristojorg, thank you for this PR. it looks good, it is also backward compatible, I might make some more changes, as the declarations seem to be duplicated in dist and also into dist/src. I'll merge this into develop and make my changes on top of it

aferditamuriqi (approved):

kristojorg:
Yea I noticed that too. One thing you could do is to add a `"rootDir": "./src"` option to the `tsconfig`, which will mean there will no longer be a `dist/src` folder, but the contents of `src` will go directly into `dist`. You might then have to explicitly include the folders outside dist, like `injectables` in the tsconfig `include` property, or move them into src.

PR #153, opened 2021-03-23
Small typefix

kristojorg:
Here is a small change to the `Injectable` types. I am not _certain_ that it is meant to be this way, but it appears so based on the examples not always having these values defined. I was getting a type error from not defining them myself. Just a small thing, but figured I'd make a quick PR in case it's helpful. Feel free to reject if not

### DIFF
--- a/src/navigator/IFrameNavigator.ts
+++ b/src/navigator/IFrameNavigator.ts
@@ -80,9 +80,9 @@ export interface IFrameNavigatorConfig {
 export interface Injectable {
     type: string;
     url?: string;
-    r2after: boolean;
-    r2before: boolean;
-    r2default: boolean;
+    r2after?: boolean;
+    r2before?: boolean;
+    r2default?: boolean;
     fontFamily?: string;
     systemFont?: boolean;
     appearance?: string;


kristojorg:
On a related note, I had to include all of these properties in the `D2Reader.load` call in order to satisfy the typings. I assume most of them should actually be optional? 
```js
const reader = await D2Reader.load({
      url,
      injectables: injectables as any,
      api: {
        getContent,
      },
      // all of these were required
      userSettings: undefined,
      initialAnnotations: undefined,
      lastReadingPosition: undefined,
      upLinkUrl: undefined,
      material: {
        settings: {
          fontOverride: false,
          advancedSettings: false,
          pageMargins: false,
          lineHeight: false,
        },
      },
      rights: {},
      tts: undefined,
      search: { color: 'red', current: 'blah' },
      annotations: { initialAnnotationColor: 'blue' },
      highlighter: { selectionMenuItems: [] },
      useLocalStorage: false,
      attributes: { margin: 2 },
    });
```

aferditamuriqi:
Yes all of them should be optional, looks like now that we enabled typings they are not treated as optional. I'll look into this.

aferditamuriqi (approved):

PR #165, opened 2021-04-12
fixed vertical scroll, removed scrollvalues

aferditamuriqi:


src/model/user-settings/UserSettings.ts
```diff
@@ -111,14 +111,14 @@ export class UserSettings implements IUserSettings {
     let scroll =
       (await this.getProperty(ReadiumCSS.SCROLL_KEY)) != null
         ? ((await this.getProperty(ReadiumCSS.SCROLL_KEY)) as Switchable).value
-        : this.verticalScroll;
+        : UserSettings.parseScrollSetting(this.verticalScroll);
```
kristojorg: I think `parseScrollSetting` should only be needed when you're taking one of the input settings and transforming it into a boolean. I believe `this.verticalScroll` should always already be a boolean though. Leaving this as `this.verticalScroll` should work, along with at line 121 below

src/model/user-settings/UserSettings.ts
```diff
@@ -821,13 +821,14 @@ export class UserSettings implements IUserSettings {
...
-            this.userProperties.getByRef(ReadiumCSS.SCROLL_REF).value = index;
+            this.userProperties.getByRef(ReadiumCSS.SCROLL_REF).value =
+              index == 0 ? true : false;
             this.storeProperty(
               this.userProperties.getByRef(ReadiumCSS.SCROLL_REF)
             );
             this.applyProperties();
             this.updateViewButtons();
-            this.view.setMode(index === 0);
+            this.view.setMode(index == 0 ? true : false);
```
kristojorg: In general I think we should use `===` instead of `==` since `==` does type coercion in sometimes unexpected ways. For example:
```
false == 0 // true
"NaN" == 0 // true
undefined == 0 // true

false === 0 // false
"NaN" === 0 // false
undefined === 0 // false
0 === 0 // true
```

src/model/user-settings/UserSettings.ts
```diff
@@ -868,15 +869,16 @@ export class UserSettings implements IUserSettings {
           index
         ].className.replace(" active", "");
       }
+
+      const index =
+        (await this.userProperties.getByRef(ReadiumCSS.SCROLL_REF).value) ==
```
kristojorg: Same comment about `==` here

kristojorg (commented):
Looks good, just a couple comments : ). Thanks for helping with this. I think it'll make the code clearer to have it modeled as a regular boolean

src/model/user-settings/UserSettings.ts
aferditamuriqi: yeah that's what i thought as well, but for some reason it didn't work, i'll try again though

src/model/user-settings/UserSettings.ts
aferditamuriqi: you're right, i already changed that in some other places.

src/model/user-settings/UserSettings.ts
aferditamuriqi: yes that works, reverting back.

PR #166, opened 2021-04-13
applied lint suggested code quality

aferditamuriqi:


src/modules/highlight/TextHighlighter.ts
```diff
@@ -817,7 +817,7 @@ export default class TextHighlighter {
           endOffset = selection.focusOffset;
         selection.collapse(selection.anchorNode, selection.anchorOffset);
 
-        var direction = [];
+        var direction: any[];
```
kristojorg: I think declaring this as `let direction: string[] = [];` is the safest way to do this. Otherwise if you don't instantiate it, Typescript thinks it is definitely an array, but it isn't actually an array until it's instantiated... so if in the future we forget to instantiate it or something we'll run in to problems.

src/modules/highlight/common/events.ts
```diff
@@ -19,152 +19,25 @@
...
-export interface IEventPayload_R2_EVENT_PAGE_TURN {
-  direction: string;
-  go: string;
-}
-
-// in WEBVIEW: ipcRenderer.sendToHost()
-// in RENDERER: webview.addEventListener("ipc-message")
-export const R2_EVENT_READING_LOCATION = "R2_EVENT_READING_LOCATION";
-
-// tslint:disable-next-line:class-name
```
kristojorg, line 53: This looks like a pretty big change. Not sure if it was intended or not, but just pointing it out in case it wasn't supposed to be here. If it was unused then feel free to ignore me : )

src/modules/highlight/renderer/common/cssselector.ts
```diff
@@ -27,11 +27,11 @@ export const fullQualifiedSelector = (
...
 ): string => {
   if (node.nodeType !== Node.ELEMENT_NODE) {
-    const lowerCaseName =
+    return (
       ((node as Element).localName &&
         (node as Element).localName.toLowerCase()) ||
-      node.nodeName.toLowerCase();
-    return lowerCaseName;
+      node.nodeName.toLowerCase()
+    );
```
kristojorg: One way to write this that might be easier to read is:
```
return (node as Element)?.localName.toLowerCase() ?? node.nodeName.toLowerCase()
```
where you use optional chaining plus [nullish coalescing](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html) to provide a fallback in case it ends up being undefined somewhere

kristojorg (approved):
Wow you made quick work of those lint suggestions 💪 ! Looks great to me, just left a couple of ideas in there in case you want to incorporate them

src/modules/highlight/TextHighlighter.ts
kristojorg: For example I'm pretty sure typescript won't warn you about this:
```
var direction: any[];
direction.push("hi"); // ERROR because direction is still undefined, even though TS thinks it's an array
```

aferditamuriqi: i guess this happens when suggested autocomplete is accepted :) i'll change it tomorrow

src/modules/highlight/common/events.ts
aferditamuriqi: none of this was actually used

src/modules/highlight/renderer/common/cssselector.ts
aferditamuriqi: I'll double check this as well tomorrow, this was also a result of accepting suggested lint ;)

src/modules/highlight/TextHighlighter.ts
aferditamuriqi: actually in this scenario, we never use it to push any values, since it's always initiated in the next few lines

aferditamuriqi: i could do this instead i guess:
```
        let direction = ["forward", "backward"];
        if (backwards) {
          direction = ["backward", "forward"];
        }

```

src/modules/highlight/renderer/common/cssselector.ts
aferditamuriqi: i'll use your recommendation, this entire function is currently not even used, but i didn't want to delete it yet before i double check if we need it

PR #167, opened 2021-04-14
Remove option chain

kristojorg:
This is a PR to remove the ts-optchain package, and refactor any code using it to instead rely on the regular typescript optional chaining and nullish coalescing. 

I did this using an automated codemod, and looked over the changes after. I will do a full review on this PR and I think we should at least do some testing as there is a chance that the codemod messed up some edge cases, though I tried to handle them.

We could either merge this now before testing is finished on the eslint-prettier branch or wait until it is merged to develop and merge it in there after. Either way, putting this here so we can start reviewing it.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -58,23 +58,23 @@ export async function unload() {
   document.body.onscroll = () => {};
   R2Navigator.stop();
   R2Settings.stop();
-  if (oc(R2Navigator.rights).enableTTS(false)) {
+  if (R2Navigator.rights?.enableTTS) {
     R2TTSSettings.stop();
     TTSModuleInstance.stop();
   }
-  if (oc(R2Navigator.rights).enableBookmarks(false)) {
+  if (R2Navigator.rights?.enableBookmarks) {
     BookmarkModuleInstance.stop();
   }
-  if (oc(R2Navigator.rights).enableAnnotations(false)) {
+  if (R2Navigator.rights?.enableAnnotations) {
     AnnotationModuleInstance.stop();
   }
-  if (oc(R2Navigator.rights).enableSearch(false)) {
+  if (R2Navigator.rights?.enableSearch) {
     SearchModuleInstance.stop();
   }
-  if (oc(R2Navigator.rights).enableContentProtection(false)) {
+  if (R2Navigator.rights?.enableContentProtection) {
     ContentProtectionModuleInstance.stop();
   }
-  if (oc(R2Navigator.rights).enableTimeline(false)) {
+  if (R2Navigator.rights?.enableTimeline) {
     TimelineModuleInstance.stop();
   }
 }
@@ -104,7 +104,7 @@ export function resumeReadAloud() {
 }
 
 export async function saveBookmark() {
-  if (oc(R2Navigator.rights).enableBookmarks(false)) {
+  if (R2Navigator.rights?.enableBookmarks) {
     if (IS_DEV) {
       console.log("saveBookmark");
     }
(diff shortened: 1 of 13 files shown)


kristojorg:
For reference and in case we need it for the future, here is the codemod I ran with `jscodeshift`:
```js
export default (fileInfo, api) => {
  const j = api.jscodeshift;
  const root = j(fileInfo.source);

  const ocShape = {
    type: "MemberExpression",
    object: {
      type: "CallExpression",
      callee: {
        type: "Identifier",
        name: "oc"
      }
    }
  };

  // first do the plain if statements with a default boolean of false
  const ifCalls = root.find(j.IfStatement, {
    test: {
      type: "CallExpression",
      callee: ocShape
    }
  });

  ifCalls.replaceWith((nodePath) => {
    const { node } = nodePath;
    const innerMember = node.test.callee.object.arguments[0];
    const accessedValue = node.test.callee.property;
    const body = node.consequent;

    // don't modify if the default isn't false
    if (!(
      node.test.arguments[0] &&
      node.test.arguments[0].type === "BooleanLiteral" 
      && node.test.arguments[0].value === false
      )
    ) {
      return node;
    }

    return {
      ...node,
      test: {
        type: "OptionalMemberExpression",
        optional: true,
        object: innerMember,
        property: accessedValue
      }
    };
  });

  const ocCalls = root.find(j.CallExpression, {
    callee: ocShape
  });

  ocCalls.replaceWith((nodePath) => {
    const { node } = nodePath;
    const defaultValue = node.arguments[0];

    const innerMember = node.callee.object.arguments[0];
    const accessedValue = node.callee.property;

    const left = {
      type: "OptionalMemberExpression",
      optional: true,
      object: innerMember,
      property: accessedValue
    };

    // if there is no default value, we keep it as a function
    if (!defaultValue) {
      return {
        type: "OptionalCallExpression",
        callee: left,
        optional: false,
        arguments: []
      };
    }

    return {
      type: "LogicalExpression",
      operator: "??",
      extra: {
        parenthesized: true
      },
      right: defaultValue,
      left
    };
  });

  return root.toSource();
};
```

src/index.ts
```diff
@@ -262,7 +261,7 @@ export async function decrease(incremental) {
     (incremental === "pitch" ||
       incremental === "rate" ||
       incremental === "volume") &&
-    oc(R2Navigator.rights).enableTTS(false)
+    (R2Navigator.rights?.enableTTS ?? false)
```
aferditamuriqi: do we need `?? false` o should that work without it , just this `R2Navigator.rights?.enableTTS`

src/modules/highlight/TextHighlighter.ts
```diff
@@ -990,43 +990,41 @@ export default class TextHighlighter {
...
-                  element,
-                  self.delegate.iframe.contentDocument,
-                  options
-                );
-              }
-
-              const selectionInfo = getCurrentSelectionInfo(
-                self.delegate.iframe.contentWindow,
-                getCssSelector
+      if (this.config?.selectionMenuItems ?? []) {
```
aferditamuriqi, line 986: maybe this if statement is obsolete, it was before as well, since the foreach would only run if the array is not empty

kristojorg: Hmm yea maybe, but I think we should leave it for now and make that change in a follow up if we want to

aferditamuriqi: 👍

aferditamuriqi (approved):
Looks good code wise, we would need to test it, but overall this feel all good.

PR #168, opened 2021-04-15
Several fixes and refactors based on testing

aferditamuriqi:


src/model/user-settings/UserSettings.ts
```diff
@@ -40,7 +40,7 @@ export interface UserSettingsConfig {
   initialUserSettings: InitialUserSettings;
   headerMenu: HTMLElement;
   material: ReaderUI;
-  api: any;
+  api: NavigatorAPI;
```
kristojorg, line 43: 👍

src/modules/AnnotationModule.ts
```diff
@@ -43,20 +43,23 @@ export interface AnnotationModuleAPI {
   deleteAnnotation: Highlight;
   selectedAnnotation: Highlight;
 }
-
 export interface AnnotationModuleConfig {
+  initialAnnotationColor: string;
+  api: AnnotationModuleAPI;
```
kristojorg: Question - I don't really know this module very well, but I'm wondering why the `api` would be nested in the `config`? What is the config meant to represent exactly? To me it would seem like the settings that get passed in when setting up/creating an `AnnotationModule`, but it looks like that is the `AnnotationModuleProperties`. It's definitely not something we need to change, especially in this PR, but I guess when I look further down you often reference `this.config.api` whereas it would seem more straightforward to just reference `this.api`? Idk I guess I'm confused on the role of `config`

src/modules/highlight/TextHighlighter.ts
```diff
@@ -1876,8 +1888,13 @@ export default class TextHighlighter {
...
         // if(anno.comment) {
-        if (this.delegate.api && this.delegate.api.selectedAnnotation) {
-          this.delegate.api.selectedAnnotation(anno).then(async () => {});
+        if (
+          this.delegate.annotationModule.config.api &&
+          this.delegate.annotationModule.config.api.selectedAnnotation
+        ) {
+          this.delegate.annotationModule.config.api
+            .selectedAnnotation(anno)
+            .then(async () => {});
```
kristojorg: This can be changed to:
```ts
this.delegate.annotationModule.config.api?.selectedAnnotation(anno)
```
I believe, but it's not a huge deal either way

kristojorg (approved):
This all looks good to me! I don't totally understand all these modules, so I can't say I totally understand the changes, but most of them look pretty straightforward and I like that a lot of `any` times are being replaced with something concrete!

It looks like you call the argument used to instantiate these classes the `properties`. I personally would probably call that the `config` and the internal object the `properties` or `settings` or something, but it's just personal preference. I would also pull out things that aren't just settings, like the `api` and store that on the instance itself so that you can call `this.api` instead of `this.config.api`, but again not a big deal. I also don't know this code well so might be totally misreading this : ). Anyways, just a comment, feel free to ignore!

src/modules/highlight/TextHighlighter.ts
aferditamuriqi: yes, this should be now possible, just didn't want to refactor in the first step  too much.

src/modules/AnnotationModule.ts
aferditamuriqi: yes, i can explain that:

when loading the reader, you have for each module a section, like the tts section, which has it's own api section, as well as other configuration elements like autoScroll, color etc. 
when using the color in the tts module it would be this.config.initialAnnotationColor, and the api this.config.api. and this would be across all modules, 
on the higher level the api is for the navigator itself. 
```
D2Reader.load({
            tts: {
                enableSplitter: true,
                color: "orange",
                autoScroll: true,
                api: {
                    started: function () {
                    },
                    stopped: function () {
                    },
                    paused: function () {
                    },
                    resumed: function () {
                    },
                    finished: function () {
                    },
                    updateSettings: function (settings) {
                    }
                }
            },
            search: {
                color: "#fce300",
                current: "#ed7138",
                api: {
                }
            },
            bookmarks: {
                api: {
                    addBookmark: function (bookmark) {
                    },
                    deleteBookmark: function (bookmark) {
                    }
                }
            },
            api: {
                getContent: function (href) {
                },
                updateCurrentLocation: function (location) {
                },
                updateSettings: function (settings) {
                }
            },
            injectables: injectables,
            userSettings: userSettings,
            useLocalStorage: true // true = uses local storage, false = uses session storage (default)
        })

src/modules/highlight/TextHighlighter.ts
aferditamuriqi: i will make this change in a separate iteration, since this is used in more than one place, and more than one module

src/modules/AnnotationModule.ts
aferditamuriqi: but i sure see you r point about renaming config, properties etc.

kristojorg: Ok I see what you're saying... there is the `api` of the navigator, and then the `api` of the individual modules, and that they need to be distinct. I think this all makes sense now. The one thing that I think would make it simpler maybe is to just get rid of the `AnnotationsModuleProperties` interface since it's not really needed. The `AnnotationsModule` is the only thing using it, and you already have the properties typed directly on the class itself, so it's not really necessary. Same can be said of the other `XXXModuleProperties` interfaces. Up to you though, it all makes sense to me now as is anyway

aferditamuriqi: does this make more sense ? https://github.com/d-i-t-a/R2D2BC/pull/170

PR #187, opened 2021-05-27
whitespace-span fixes

aferditamuriqi:
this should fix the issue with the white space spans, instead of adding just a text white space. which should not cause any issues.

### DIFF
--- a/src/modules/TTS/splitting.js
+++ b/src/modules/TTS/splitting.js
@@ -248,7 +248,10 @@
 
       // If there's no text left after trimming whitespace, continue the loop
       if (contentsTrimmed.length) {
-        allElements.push(createText(" "));
+        // insert leading space if there was one
+        if (wholeText[0] === " ") {
+          allElements.push(createText(" "));
+        }
         // Concatenate the split text children back into the full array
         each(contentsTrimmed.split(splitOn), function (splitText, i) {
           if (i && preserveWhitespace) {
@@ -260,8 +263,12 @@
           elements.push(splitEl);
           allElements.push(splitEl);
         });
-        allElements.push(createText(" "));
-      } else {
+        // insert trailing space if there was one
+        if (wholeText[wholeText.length - 1] === " ") {
+          allElements.push(createText(" "));
+        }
+      } else if (wholeText.length && !contentsTrimmed.length) {
+        // add back missing whitespace
         allElements.push(createText(" "));
       }
     });
--- a/src/modules/TTS/splitting.js
+++ b/src/modules/TTS/splitting.js
@@ -244,6 +244,7 @@
       // Get the text to split, trimming out the whitespace
       /** @type {string} */
       var wholeText = next.wholeText || "";
+      wholeText = wholeText.replace(/\s+/g, " ");
       var contentsTrimmed = wholeText.trim();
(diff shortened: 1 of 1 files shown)


src/modules/TTS/splitting.js
```diff
@@ -245,17 +245,10 @@
...
       // If there's no text left after trimming whitespace, continue the loop
       if (contentsTrimmed.length) {
-        // insert leading space if there was one
-        allElements.push(
-          createElement(F, "whitespace", " ", preserveWhitespace)
-        );
-        if (wholeText[0] === " ") {
-          allElements.push(createText(" "));
-        }
+        allElements.push(createText(" "));
```
mbrambilla: This is adding spaces where ones previously did not exist.   Consider using previous whitespace check before pushing space:
```
// insert leading space if there was one
if (wholeText[0] === " ") {
    allElements.push(createText(" "));
}
```

src/modules/TTS/splitting.js
```diff
@@ -267,21 +260,9 @@
...
           allElements.push(splitEl);
         });
-        // insert trailing space if there was one
-        if (wholeText[wholeText.length - 1] === " ") {
-          allElements.push(createText(" "));
-        }
-        allElements.push(
-          createElement(F, "whitespace", " ", preserveWhitespace)
-        );
+        allElements.push(createText(" "));
```
mbrambilla: Again potentially adding space where one did not exist before.  Use previous check again:
```
// insert trailing space if there was one
if (wholeText[wholeText.length - 1] === " ") {
    allElements.push(createText(" "));
}
```

src/modules/TTS/splitting.js
```diff
@@ -267,21 +260,9 @@
...
       } else {
-        // Don't create whitespace element where textNode is invalid
-        var isChildTextNodeInvalid = /table|thead|tbody|tfoot|tr/i.test(el.tagName);
-        if (!isChildTextNodeInvalid) {
-          allElements.push(
-            createElement(F, "whitespace", " ", preserveWhitespace)
-          );
-        }
+        allElements.push(createText(" "));
```
mbrambilla, line 273: Maybe conditional else if check here to add space only when needed, instead of just else.
```
} else if (wholeText.length && !contentsTrimmed.length) {
    // add back missing whitespace
    allElements.push(createText(" "));
}
```

src/modules/TTS/splitting.js
aferditamuriqi: reverted this

src/modules/TTS/splitting.js
aferditamuriqi: reverted this.

src/modules/TTS/splitting.js
aferditamuriqi: reverted this.

src/modules/TTS/splitting.js
aferditamuriqi: added suggested else if

mbrambilla (approved):
This looks good to me.  All the test cases I can come up with are coming out as expected.

Seems like there is a minor improvement in render speed too, but that could be circumstantial.

PR #196, opened 2021-06-14
Make currentSettings not async

kristojorg:
This simply removes the `async` keyword from the `currentSettings` function, and the subsequent file in `UserSettings`. The keyword added an unnecessary promise wrapper to the function response which made reading the settings for display in a react app somewhat complicated. 

I also made `R2Settings.currentSettings` a getter, so it can now be accessed as a static property instead of as a function.

Consumers will need to update their code to no longer expect a promise here.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -238,7 +238,7 @@ export async function applyUserSettings(userSettings) {
   }
   R2Settings.applyUserSettings(userSettings);
 }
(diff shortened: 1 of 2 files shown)


kristojorg:
I am actually going to close this in favor of a proposed v2 api into the reader using a single class...

PR #197, opened 2021-06-14
New Class-based API

kristojorg:
This PR introduces a new `Reader` class which is exported from `index.ts`. It is meant to be a replacement for the other API exposed in `index.ts`. The motivation behind doing this is that with the old method, `vars` declared at the top of the file were not guaranteed to be instantiated by the time you called functions that depended on them. This was causing me a lot of headaches as I couldn't find a good way to know when the methods were ready to be used. 

To resolve it, I export a single class with an async `build` function. Once you `build` an instance, it is completely ready and all the methods can be accessed. So you can simple `const reader = await Reader.build(config)` and then safely call `reader.nextPage()` with the instance you now have. 

There were a couple other changes that I had to make in order for this to work:
- I briefly turned on `strictNullChecks` and had to mark many items in different modules as optional.
- I added a [Parcel](https://parceljs.org) based react example using the new class interface. It watches the files and rebuilds automatically and quickly on any change so you can see them immediately in the browser. 
- In order for the parcel example to work, I moved the `tsconfig` to the project root and modified it a bit.
- I also removed the `async` wrapper from `currentSettings` so that it immediately returns the result.

## New Example App

To run the example: `npm run example:reader-class`. You will note that the `nextPage` and `previousPage` buttons do work, but the `scroll` button does not seem to update to `paginate` once you are in scrolling mode. This appears to be because `R2Settings.currentSettings` does not reflect the current value actually being used.

## To Do

- [x] Add the code from `load` that updates the passed in `config` value before starting the modules.
- [x] Add function to ensure browser support before starting?
- [x] Type `Reader` class method arguments

## Next Steps

Transitioning to an API like this should make it easier to turn on `strictNullChecks`, though we could still support the old API if we want.


src/Reader.ts
```diff
@@ -0,0 +1,442 @@
...
+    readonly ttsModule?: TTSModule,
+    readonly searchModule?: SearchModule,
+    readonly contentProtectionModule?: ContentProtectionModule,
+    readonly timelineModule?: TimelineModule
+  ) {}
+  /**
+   * The async builder.
+   * @TODO :
+   *  - browser support section
+   *  - Code that updates/changes the config before we start using it
```
kristojorg: These are two big things from the old `load` function in `index.ts` that I have yet to build into this.

kristojorg:
Ok I made a few more updates here:

1. I moved the code to enforce supported browsers into a util inside `utils/BrowserUtilities.ts`. It is now being used inside `Reader.ts`, but I have not made any changes to `index.ts`. We can replace that code with this util if it looks good.
2. I moved the code for generating or fetching positions and weights into the `Publication` class. It is now being used inside `Reader.ts` but is not being used in `index.ts` yet. Again, we can replace that code with the refactored version if it looks good to you.
3. I left some comments on things I'm a bit confused on.

Before we merge this, I would also like to type all function arguments in the `Reader` class. Maybe you can help me with that @aferditamuriqi ?

src/model/Publication.ts
```diff
@@ -247,4 +247,132 @@ export default class Publication {
...
+    const positions = [];
+
+    /**
+     * For each item in the reading order, get its length and calculate
+     * the number of positions in it, then add them all up into the totals.
+     */
+    const promises = this.readingOrder.map(async (link, index) => {
+      // if it is fixed layout, there is no need to fetch, each item is
+      // just a single page.
+      // @aferditamuriqi is this comment correct?
```
kristojorg: LMK if this is right?

src/model/Publication.ts
```diff
@@ -247,4 +247,132 @@ export default class Publication {
...
+    // @aferditamuriqi but what is the "weight" object ultimately used for?
+    const weight = {};
+    if (this.metadata.rendition?.layout !== "fixed") {
+      this.readingOrder.forEach((link) => {
+        // I don't totally know what this formula is saying haha, maybe we can
+        // add a comment
+        link.contentWeight = (100 / totalContentLength) * link.contentLength;
+        weight[link.href] = link.contentWeight;
+      });
+    }
```
kristojorg: I pulled this out from inside the loop (you can see the old commented out version still in there). LMK if this looks okay? I couldn't really understand how the `weight` variable was used in the old `index.ts` code. Maybe you can help me out there? Would also be helpful to have a little comment on this block to know what it's meant to do.

src/model/Publication.ts
```diff
@@ -247,4 +247,132 @@ export default class Publication {
...
+          );
+          locator.locations.totalProgression =
+            (locator.locations.position - 1) / positions.length;
+          locator.locations.remainingPositions = Math.abs(
+            positionIndex - (resource.length - 1)
+          );
+          locator.locations.totalRemainingPositions = Math.abs(
+            locator.locations.position - 1 - (positions.length - 1)
+          );
+        });
```
kristojorg: This part I don't totally understand. Can you tell me what it's meant to do and if it can be pulled out of the above loop?

src/model/Publication.ts
```diff
@@ -247,4 +247,133 @@ export default class Publication {
...
+        link.contentWeight = (100 / totalContentLength) * link.contentLength;
+        weight[link.href] = link.contentWeight;
+      });
+    }
+
+    // we need to wait for all of them to complete, meaning everything has bee
+    // fetched and counted
+    await Promise.all(promises);
+
+    this.positions = positions;
```
kristojorg: I also pulled this out of the loop and am waiting for all the promises generated by our fetches to resolve before setting `this.positions`.

src/model/Publication.ts
```diff
@@ -247,4 +247,133 @@ export default class Publication {
...
+              position: startPosition + (position + 1),
+            },
+            type: link.type,
+          };
+          positions.push(locator);
+        });
+        startPosition = startPosition + positionCount;
+      }
+
+      // If we are on the last item...
```
kristojorg: Instead of testing whether we are on the last item, should this content just be pulled out to run after the `map` loop is finished?

.gitignore
```diff
@@ -9,8 +9,9 @@ reader.js
 reader.js.map
 examples/streamed/readers/viewer
 viewer/injectables
-viewer/index.js
-viewer/index.js.map
+viewer/index*.js
+viewer/index*.js.map
```
kristojorg: This should just be index.js. Maybe because of the ts config change?

kristojorg: Injectables should be separate

src/Reader.ts
```diff
@@ -0,0 +1,485 @@
...
+   */
+  static async build(initialConfig: ReaderConfig): Promise<Reader> {
+    // will throw error if on unsupported browser
+    enforceSupportedBrowsers(initialConfig);
+
+    const mainElement = getElement("D2Reader-Container");
+    // are the following elements necessary or not? They seem not to be,
+    // but we will have to change some types if they are allowed to be null
+    const headerMenu = getElement("headerMenu");
+    const footerMenu = getElement("footerMenu");
```
kristojorg: Replace `getElement` with `findElement` or `findRequiredElement` from `HTMLUtilities`

kristojorg: Header menu and footer menu are no longer required. Just update their types

kristojorg: Refactor these out into a list of reserved ids

src/Reader.ts
```diff
@@ -0,0 +1,485 @@
...
+
+    const upLink: UpLinkConfig = initialConfig.upLinkUrl ?? undefined;
+
+    const publication: Publication = await Publication.getManifest(
+      webpubManifestUrl,
+      store
+    );
+
+    // update our config based on what we know from the publication
+    const config = updateConfig(initialConfig, publication);
```
kristojorg: Rename this to be specific to fixed layout

src/Reader.ts
```diff
@@ -0,0 +1,485 @@
...
+    if (config.rights?.autoGeneratePositions ?? true) {
+      await publication.autoGeneratePositions();
+    } else {
+      if (config.services?.positions) {
+        await publication.fetchPositionsFromService(
+          config.services?.positions.href
+        );
+      }
+      if (config.services?.weight) {
+        await publication.fetchWeightsFromService(config.services?.weight.href);
```
kristojorg: Weight is so that each link knows how large it is relative to the total

src/Reader.ts
```diff
@@ -0,0 +1,485 @@
...
+  };
+  currentSettings = () => {
+    return this.settings.currentSettings();
+  };
+  scroll = async (value) => {
+    return await this.settings.scroll(value);
+  };
+
+  /**
+   * pitch?
```
kristojorg: To increase anything that can be increased, such as `fontSize`

src/Reader.ts
```diff
@@ -0,0 +1,485 @@
...
+      this.navigator.rights?.enableTTS
+    ) {
+      this.ttsSettings.decrease(incremental);
+    } else {
+      this.settings.decrease(incremental);
+    }
+  };
+
+  /**
+   * Publisher?
```
kristojorg: The opposite of advanced settings on... if the advanced settings are off, we use the default publisher settings.

kristojorg: boolean

src/Reader.ts
```diff
@@ -0,0 +1,485 @@
...
+  };
+  atEnd = async () => {
+    return this.navigator.atEnd();
+  };
+  snapToElement = async (value) => {
+    this.navigator.snapToElement(value);
+  };
+
+  /**
+   * ??
```
kristojorg: You have `attributes` in the reader when you initialize it. You can set `margin`, `navigationHeight` etc... 

This is in case you change the attributes after initializing the reader.

`IframeAttribute`

src/Reader.ts
```diff
@@ -0,0 +1,485 @@
...
+   */
+  applyAttributes = (value) => {
+    this.navigator.applyAttributes(value);
+  };
+
+  /**
+   * Destructor
```
kristojorg: Only used in react applications. Because when they re-visit the page it tried to create a new reader, which interfered with the first one.

src/index.ts
```diff
@@ -35,6 +35,7 @@ import TextHighlighter from "./modules/highlight/TextHighlighter";
 import { Locator } from "./model/Locator";
 import TimelineModule from "./modules/positions/TimelineModule";
 import { getUserAgentRegExp } from "browserslist-useragent-regexp";
+export { default as Reader } from "./Reader";
```
kristojorg: Let's switch to the new one and make a document about migration

kristojorg: Name it `D2Reader`

src/model/Publication.ts
kristojorg: It's right

src/model/Publication.ts
```diff
@@ -247,4 +247,133 @@ export default class Publication {
...
+        startPosition = startPosition + 1;
+      } else {
+        const href = this.getAbsoluteHref(link.href);
+        // Are we fetching every item in the readingOrder to generate positions?
+        // maybe in the future this can be done in the background?
+        const result = await fetch(href);
+        const length = (await result.blob()).size;
+        link.contentLength = length;
+        totalContentLength += length;
+        const positionLength = 1024;
```
kristojorg, line 234: This is what we want to make configurable

src/model/Publication.ts
kristojorg: Once you have all the positions, you can update all the progressions and total progressions and remaining. This can be moved out of the loop as well

tsconfig.json
```diff
@@ -18,9 +18,10 @@
  */
 
 {
+  "include": ["src", "examples"],
+  "exclude": ["viewer"],
```
kristojorg: Injectables should not be part of the source either. They're just examples

kristojorg (commented):
Comments

src/model/Publication.ts
aferditamuriqi: yes this is correct

PR #199, opened 2021-06-15
Feature: Media Overlays and R2 models replacement

aferditamuriqi:
- replace internal models with r2-shared-js models
- refactored to use array of iframes (relevant for fxl)
- several fixes on user settings
- turned autogenerate positions to default false 
- upgraded example streamer 

Added Feature: Media Overlays


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@d-i-t-a/reader",
-  "version": "1.8.5",
+  "version": "2.0.0",
```
kristojorg: I think actually we should make this 2.0.0-beta.1 and keep publishing changes to the beta until we have settled a bit. I think there may be a few more breaking changes we want to make in addition to the new API, for example if we implement strictNullChecks or change some of the defaults, etc. I think it's safest to make it a beta release for a little while so we can add more breaking changes if need be without annoying the people using the stable version. That way they just upgrade a single time when we release 2.0.0

src/model/Link.ts
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * Developed on behalf of: DITA
+ * Licensed to: Bokbasen AS and CAST under one or more contributor license agreements.
+ */
+
+import { Link as R2Link } from "r2-shared-js/dist/es6-es2015/src/models/publication-link";
+import { JsonObject } from "ta-json-x";
+
+export class D2Link {
+  href?: string;
```
kristojorg, line 24: Is the href supposed to be optional here?

src/navigator/IFrameNavigator.ts
```diff
@@ -159,15 +160,16 @@ export interface ReaderConfig {
...
   useLocalStorage?: boolean;
   attributes?: IFrameAttributes;
   services?: PublicationServices;
 }
 
 /** Class that shows webpub resources in an iframe, with navigation controls outside the iframe. */
 export default class IFrameNavigator implements Navigator {
-  iframe: HTMLIFrameElement;
-  iframe2: HTMLIFrameElement;
+  iframes: Array<HTMLIFrameElement> = [];
```
kristojorg, line 177: Is this always supposed to have exactly two iframes in it? If so it might make sense to type it as a tuple: 
```ts
iframes: [HTMLIFrameElement, HTMLIFrameElement];
```
that way Typescript know that there are always exactly two items in it, and will keep us from trying to access `iframes[2]` or something, and we can also be sure that `iframes[0]` and `iframes[1]` will always be defined.

src/navigator/IFrameNavigator.ts
```diff
@@ -1799,65 +1894,106 @@ export default class IFrameNavigator implements Navigator {
...
+            } else {
+              fetch(this.currentChapterLink.href)
+                .then((r) => r.text())
+                .then(async (content) => {
+                  writeIframeDoc.call(
+                    this,
+                    content,
+                    this.currentChapterLink.href
+                  );
+                });
```
kristojorg, line 1946: Doesn't need to be changed here, but just so you know, this can be rewritten with async/await as:
```ts
const r = await fetch(this.currentChapterLink.href)
const content = await r.text();
writeIframeDoc.call(this, content, this.currentChapterLink.href)
```
which avoids then nesting of `.then` calls. Basically anything you would call `something().then(value => {...})` you can use `const value = await something()` so long as you're in an `async` function

kristojorg (commented):
Just a couple comments

package.json
kristojorg: We discussed this in slack that we are going to use 1.9.0 for this release

src/model/Link.ts
aferditamuriqi: yes this is actually just a helper class, i wanted to avoid having to refactor it all, there are several places where a link is constructed from nothing, and only needs the href , title and type.  could be refactored later on as needed.

package.json
aferditamuriqi: yes i will make sure to change this to 1.9

src/navigator/IFrameNavigator.ts
aferditamuriqi: but looks much better :) thank you, will change that

src/navigator/IFrameNavigator.ts
aferditamuriqi: at some point there might be 3 iframes, there are fixed layout epubs that have 3, but for now we will either have 1 or two, in fixed layout you can have spreads of 1 to 3 i believe, but we are just using 1 or 2 here

aferditamuriqi: i am using the user-settings column count as page count here, where the user could set the book to show in one page vs two, or vise versa. some fixed layout books are not laid out properly and have two pages in one spread, so it's good to be able to only show one in that case.

src/navigator/IFrameNavigator.ts
aferditamuriqi: actually i will leave it for now as is, since it seems to require more changes to make it work.

src/model/user-settings/UserSettings.ts
```diff
@@ -446,9 +450,8 @@ export class UserSettings implements IUserSettings {
         "#container-view-settings"
       ) as HTMLDivElement;
     if ((await this.getProperty(ReadiumCSS.SCROLL_KEY)) != null) {
-      (
-        await this.getProperty(ReadiumCSS.SCROLL_KEY)
-      ).value = this.verticalScroll;
+      (await this.getProperty(ReadiumCSS.SCROLL_KEY)).value =
+        this.verticalScroll;
```
kristojorg: I'm not totally sure what this is supposed to do. You await the value and then set it to `this.verticalScroll`, but I don't think that will change the value in the store will it?

aferditamuriqi: i'll double check that

aferditamuriqi: you are right, good catch, that doesn't do anything. i'll change it to this:

```  
    if ((await this.getProperty(ReadiumCSS.SCROLL_KEY)) == null) {
      await this.saveProperty(
        new Switchable(
          "readium-scroll-on",
          "readium-scroll-off",
          this.verticalScroll,
          ReadiumCSS.SCROLL_REF,
          ReadiumCSS.SCROLL_KEY
        )
      );
    }
```
that should work.

package.json
aferditamuriqi: the 2.0.0 then will anyay stay in alpha and then beta for a while.

src/model/user-settings/UserSettings.ts
aferditamuriqi: removed this altogether

PR #207, opened 2021-06-22
get property update

kristojorg:
This is a small refactor that I think improves the readability around getting properties. Let me know what you think!


aferditamuriqi (approved):
~~looks good, i'll merge and test~~
actually the checks are not passing..

PR #213, opened 2021-06-29
several alpha fixes

aferditamuriqi:


src/index.ts
```diff
@@ -259,6 +284,9 @@ export default class D2Reader {
   /**
    * Read Aloud
    */
+  hasMediaOverlays = () => {
+    return this.navigator.hasMediaOverlays;
+  };
```
kristojorg, line 289: Can we make this a getter so that it can be accessed as a property?

src/index.ts
```diff
@@ -286,23 +314,19 @@ export default class D2Reader {
     }
   };
   deleteAnnotation = async (highlight: Annotation) => {
-    return await this.annotationModule?.deleteAnnotation(highlight);
+    return this.annotationModule?.deleteAnnotation(highlight);
   };
   addAnnotation = async (highlight: Annotation) => {
-    return await this.annotationModule?.addAnnotation(highlight);
```
kristojorg, line 292: The `deleteAnnotation` method is marked as async and returns a promise, so shouldn't we `await` that promise here?

src/index.ts
```diff
@@ -286,23 +314,19 @@ export default class D2Reader {
...
+    return this.annotationModule?.addAnnotation(highlight);
   };
   tableOfContents = async () => {
-    return await this.navigator.tableOfContents();
+    return await convertAndCamel(this.navigator.tableOfContents());
   };
   readingOrder = async () => {
-    return await this.navigator.readingOrder();
+    return await convertAndCamel(this.navigator.readingOrder());
```
kristojorg, line 326: The `await` here doesn't actually seem to do anything on the other hand, because `tableOfContents()` returns a value, not a promise. So I think we can make both these functions synchronous and get rid of the `async` marker in their definition. What do you think

src/index.ts
```diff
@@ -313,7 +337,7 @@ export default class D2Reader {
    */
   search = async (term: string, current: boolean) => {
     if (this.navigator.rights?.enableSearch) {
-      return await this.searchModule?.search(term, current);
```
kristojorg, line 316: This `search` method returns a promise, so we should `await` that before returning

src/index.ts
```diff
@@ -328,9 +352,9 @@ export default class D2Reader {
       await this.searchModule?.goToSearchID(href, index, current);
     }
   };
-  clearSearch = () => {
+  clearSearch = async () => {
     if (this.navigator.rights?.enableSearch) {
-      this.searchModule?.clearSearch();
+      await this.searchModule?.clearSearch();
```
kristojorg, line 357: `clearSearch` does not return a promise, however, so this `async` wrapper and `await` keyword will do nothing but wrap the whole method in a promise that resolves immediately. I think we should remove the `async` and `await` here

src/index.ts
```diff
@@ -439,31 +479,32 @@ export default class D2Reader {
...
   };
-  atEnd = async () => {
+  atEnd = () => {
     return this.navigator.atEnd();
   };
-  snapToElement = async (value: HTMLElement) => {
+  // currently not used or functional
+  snapToElement = (value: HTMLElement) => {
     this.navigator.snapToElement(value);
   };
```
kristojorg, line 509: Let's just remove it if it's not functional I think

src/index.ts
```diff
@@ -439,31 +479,32 @@ export default class D2Reader {
...
     this.navigator.previousPage();
   };
-  atStart = async () => {
+  atStart = () => {
     return this.navigator.atStart();
   };
-  atEnd = async () => {
+  atEnd = () => {
     return this.navigator.atEnd();
   };
```
kristojorg, line 505: Personally I think these can be getters

src/model/Publication.ts
```diff
@@ -43,7 +43,7 @@ export default class Publication extends R2Publication {
     return publication;
   }
 
-  get readingOrder(): Link[] {
```
kristojorg, line 46: Why remove this type?

src/model/Publication.ts
```diff
@@ -229,7 +229,7 @@ export default class Publication extends R2Publication {
         // maybe in the future this can be done in the background?
         const result = await fetch(href);
         const length = (await result.blob()).size;
-        link.contentLength = length;
+        (link as Link).contentLength = length;
```
kristojorg, line 232: Removing the `Link` type on L46 is what made it necessary to type cast this if I'm not wrong

src/model/Publication.ts
```diff
@@ -246,45 +246,47 @@ export default class Publication extends R2Publication {
...
+        // update the link.contentWeight to be a portion of the total and
+        // build up a map of link weights for non fixed layout publications
+        var totalweight = 0;
+        if (this.isReflowable) {
+          this.readingOrder.map(async (link) => {
+            // I don't totally know what this formula is saying haha, maybe we can
+            // add a comment
+            (link as Link).contentWeight =
+              (100 / totalContentLength) * (link as Link).contentLength;
+            totalweight = totalweight + (link as Link).contentWeight;
```
kristojorg, line 260: Same with these type casts

src/model/Publication.ts
```diff
@@ -246,45 +246,47 @@ export default class Publication extends R2Publication {
...
-        (locator.locations.position - 1) / positions.length;
-      locator.locations.remainingPositions = Math.abs(
-        positionIndex - (resource.length - 1)
-      );
-      locator.locations.totalRemainingPositions = Math.abs(
-        locator.locations.position - 1 - (positions.length - 1)
-      );
-      return locator;
-    });
+      if (index + 1 === this.readingOrder.length) {
```
kristojorg, line 250: Why do this while still in the `map` loop instead of after finishing the map?

src/model/Publication.ts
```diff
@@ -246,45 +246,47 @@ export default class Publication extends R2Publication {
...
+          );
+          locator.locations.totalRemainingPositions = Math.abs(
+            locator.locations.position - 1 - (positions.length - 1)
+          );
+          return locator;
+        });
+
+        this.positions = positions;
+      }
+    });
```
kristojorg, line 285: Same with this, can't this be done after the `map` loop instead of testing that we are on the last item?

src/navigator/IFrameNavigator.ts
```diff
@@ -424,6 +424,11 @@ export default class IFrameNavigator implements Navigator {
     this.headerMenu = headerMenu;
     this.mainElement = mainElement;
     try {
+      var wrapper = HTMLUtilities.findRequiredElement(
+        mainElement,
+        "main#iframe-wrapper"
+      ) as HTMLElement;
```
kristojorg, line 430: The `findRequiredElement` function now returns an `HTMLElement`, so the `as` typecasting is no longer necessary

src/navigator/IFrameNavigator.ts
```diff
@@ -1080,8 +1083,13 @@ export default class IFrameNavigator implements Navigator {
             this.isScrolling = false;
           }, 200);
 
+          const wrapper = HTMLUtilities.findRequiredElement(
+            document,
+            "#iframe-wrapper"
+          ) as HTMLDivElement;
```
kristojorg, line 1089: Same here

src/utils/BrowserUtilities.ts
```diff
@@ -19,15 +19,26 @@ import { getUserAgentRegExp } from "browserslist-useragent-regexp";
...
+import * as HTMLUtilities from "./HTMLUtilities";
 
 /** Returns the current width of the document. */
 export function getWidth(): number {
-  return document.documentElement.clientWidth;
+  const wrapper = HTMLUtilities.findRequiredElement(
+    document,
+    "#iframe-wrapper"
+  ) as HTMLDivElement;
+
```
kristojorg, line 30: We should make the `findRequiredElement` generic:
```ts
const el = findRequiredElement<HTMLDivElement>(document, "#iframe-wrapper");
```
so that we don't have to keep casting these variables

src/utils/BrowserUtilities.ts
```diff
@@ -19,15 +19,26 @@ import { getUserAgentRegExp } from "browserslist-useragent-regexp";
...
+  return wrapper.clientWidth;
 }
 
 /** Returns the current height of the document. */
 export function getHeight(): number {
-  return document.documentElement.clientHeight;
+  const wrapper = HTMLUtilities.findRequiredElement(
+    document,
+    "#iframe-wrapper"
+  ) as HTMLDivElement;
```
kristojorg, line 39: Same

kristojorg (changes requested):
So I made some comments about unnecessary `async`/`await` calls, or ones where I thought we should keep them. I guess in general we should decide if these methods should return promises or not before settling this api into 2.0.0. My thoughts are this:

- If it is possible to tell when the task is finished, we should return a promise that resolves either with a value or `void` when the task is finished. 
- If it is not possible to tell, and thus the task will appear synchronous to the caller (whether it is or isn't), we should not return a promise. 

Additionally, if we are essentially accessing a property on the `Reader` or one of its submodules, I think we should try to use getters whenever possible to simplify the code for users.

PS. Here is a good resource about the difference between `return await` and simply `return`ing another async function's promise: https://jakearchibald.com/2017/await-vs-return-vs-return-await/

I also had some thoughts about the `findRequiredElement` utility, but we could open a separate issue for that and put it in a follow up PR.

Otherwise this looks good to me

src/index.ts
aferditamuriqi: this was a suggestion form the lint, seemed like it passed on the async.

src/index.ts
aferditamuriqi: yes we could/should, we need to solve a few of my concerns regarding property s method

src/index.ts
aferditamuriqi: yes we can

src/index.ts
aferditamuriqi: this was as well a lint suggestion.

src/index.ts
aferditamuriqi: this was just brought over from how it was before, the entire search needs to be reviewed , therefore i didn't want to change too much here, since this is how it's right now in 1.9

src/model/Publication.ts
aferditamuriqi: no it couldn't, it didn't work, the positions and weight where simply empty, i've added commit notes here i beleive, why i added it back into the loop

src/navigator/IFrameNavigator.ts
aferditamuriqi: ok, i missed that, and i'll be refactoring here a bit more too, just needed to make sure the majority of the functionality still works. e.g. next / previous page

src/utils/BrowserUtilities.ts
aferditamuriqi: good idea.

src/utils/BrowserUtilities.ts
aferditamuriqi, line 39: :) same as the other response, this will be refactored.

src/index.ts
aferditamuriqi: well, i will need it soon though. i didn't want to remove it, and then forget about it, that's why the comment. we can comment it out maybe

kristojorg:
- Here is an issue for my comment about the getters: https://github.com/d-i-t-a/R2D2BC/issues/214
- Here is an issue for my comment about the `findRequiredElement` function and casting: https://github.com/d-i-t-a/R2D2BC/issues/215
- Here is an issue for my comment about the nested loops in `Publication.ts` while setting weights: https://github.com/d-i-t-a/R2D2BC/issues/216
- Issue for removing the casts to `Link` in `Publication.ts`: https://github.com/d-i-t-a/R2D2BC/issues/217

I think we can merge this without completing either of those, and put those into follow up PRs into the `v2` branch

src/index.ts
aferditamuriqi: right they can. but again this goes to the topic i want to discuss about how much we are breaking and how much we can be backwards compatible.

src/model/Publication.ts
aferditamuriqi: because this didn't actually work, not sure why, but the `contentLength` and `contentWeight` were simply always undefined

src/index.ts
kristojorg: Interesting, I also get those lint suggestions but only when the return value of the function isn't a promise... Not sure why it was doing this unless maybe it had an old version of the file in memory or something? idk

src/index.ts
kristojorg: Oh I see I thought it was just not going to be used 👍

src/model/Publication.ts
kristojorg: Gotcha, I opened a ticket to do this work later: https://github.com/d-i-t-a/R2D2BC/issues/216

src/index.ts
kristojorg: Sounds good, we can do later / discuss

src/index.ts
kristojorg: Ok sounds good

src/index.ts
kristojorg: Okay makes sense!

PR #218, opened 2021-06-30
New build system

kristojorg:
This PR replaces the Webpack build system with [Esbuild](https://esbuild.github.io). In general, the JS community has been moving away from webpack for bundling libraries because of its complexity and slowness. It tries to do everything for everyone, and therefore is pretty bad at all of it, and a pain to manage. I was having some problems with the build, and I know others were too, so I took the opportunity to make us a build setup I feel pretty good about. Some features:

## Multiple bundle types

- An ESM version at `dist/index.mjs` for use by people who import the package in bundled apps (ie people who do `import D2Reader from "@d-i-t-a/reader"`) like me.
- An IIFE version with a named export at `dist/index.js` for people who use the package via a `<script>` tag and a global variable. 

## Watch mode

The `npm run dev` command runs the build script in watch mode, rebuilding almost instantly on any change. This way when you're developing you don't have to rebuild and then test, rebuild and then test. Just refresh the browser and your new code is there.

## Local dev server

We previously would copy files from /dist into /viewer and from /viewer into /examples during a build in order to run the app locally. We now have a simple dev server that statically hosts the files in /viewer and /dist at `http://localhost:8080`.

# How does it work?

I wrote a build script at `build.ts` which does all of this. Our build pipeline looks like this:

 - clean the build folder
 - build IIFE version (for: <script href="https://../d2reader.js" />)
 - build ESM version (for: import D2Reader from "@d-i-t-a/reader")
 - build SASS (using node-sass)
 - generate TS declarations (using tsc --declarationsOnly)
 - Build IIFE version of injectables to dist
 - copy injectables css to dist

All of this is done in parallel and then we optionally will redo it whenever we detect a change. 

The nice part about having a build script is that all of our build logic now lives there instead of in npm scripts. It should give us much more control over how things are done.

# To Do

- [x] Set up dev server with the streamer...
- [ ] Update wiki / readme


kristojorg:
I think this is ready fore review. Just pushed a fix for how typescript typings were being resolved. Everything should work now. The dev server I am doing separately in another PR

src/index.ts
```diff
@@ -16,539 +16,12 @@
  * Developed on behalf of: NYPL, Bokbasen AS (https://www.bokbasen.no), CAST (http://www.cast.org)
  * Licensed to: NYPL, Bokbasen AS and CAST under one or more contributor license agreements.
  */
+import D2Reader from "./reader";
```
kristojorg, line 19: The contents of `/index.ts` were moved to `/reader.ts` so that `/index.ts` can be simply the API into the package, exporting all the things consumers need, not actually defining any implementation.

aferditamuriqi (approved):

kristojorg:
Closing in favor of #242

PR #219, opened 2021-07-02
New dev server

kristojorg:
This PR adds a `server.ts` file which is our new dev server. It uses `r2-streamer-js` to set up the express server, but also statically serves the contents of `/viewer` and `/dist` so that we don't have to manually copy files around after building.

Clicking on a publication will give you the option to view it using any of the examples in /viewer. 

We need to wait for https://github.com/readium/r2-streamer-js/pull/56 to merge and a new release before we can merge this though.

@aferditamuriqi what do you think about moving `/viewer` into `/examples/html`? That was the `/examples` folder holds all the code for the examples, whether react or html based. 

This also checks in a single example epub, but we can remove it and gitignore it if you'd rather do that. The folder for sample epubs is now: `/examples/epubs`. 

To give this a try, run `npm run build` then `npm run examples` on this branch.


kristojorg:
There hasn't been any comments on https://github.com/readium/r2-streamer-js/pull/56 so maybe we merge it into our fork of r2-streamer-js and then publish our own version of it? Or we could ask them to merge this

danielweck:
Hello, I have been thinking about it. I will get back to you shortly with comments.

kristojorg:
@danielweck thanks so much!

aferditamuriqi:
> what do you think about moving /viewer into /examples/html? That was the /examples folder holds all the code for the examples, whether react or html based.

@kristojorg  That would be fine, we just need to make sure that any of the injectable examples are still working as well. but from what i understand we will use dist then directly, for the injectables, same as the reader.js and reader.css ?

kristojorg:
Yep exactly we would be serving `dist` at `/viewer` so any page in `/viewer` can request anything within `dist`:

viewer/index.html
```html
<script src="reader.js" />
...
<link rel="stylesheet" href="reader.css" />
```

kristojorg:
How should I make sure that the injectables examples still work? How do I run them / which ones are they?

aferditamuriqi:
For example here https://github.com/d-i-t-a/R2D2BC/blob/production/viewer/index_dita.html#L554 
i could think the injectables could be compiled and placed as well into examples `/examples/injectables` and used statically from there. i wouldn't use them from dist, as these injectables are actually examples. and anyone can create their own outside of r2d2bc and inject.

aferditamuriqi:
the example injectables are specified here https://github.com/d-i-t-a/R2D2BC/blob/feature/new-streamer/webpack.config.js#L54

You see there that glossary and footnotes are really just examples, the style.css is not specified in the webpack for example.

kristojorg:
Ok right now injectables are compiled into `dist/injectables`, which is what happened with them before. I think that is the safest, but we could compile them directly into `examples/injectables` if you would rather that (I think this'll work)

kristojorg:
Closing in favor of #243

PR #220, opened 2021-07-05
Fix console.error

kristojorg:
This PR just removes a couple console.error calls that were being called very often


aferditamuriqi (approved):
right, i missed them, they shouldn't have been there :)

PR #221, opened 2021-07-05
Typescript argument fix

kristojorg:
This is a simple fix for the issue of the `applyUserSettings` argument requiring a full `UserSettings` object when in fact it should only need a `Partial<UserSettings>`


aferditamuriqi (approved):
👍

PR #222, opened 2021-07-06
Bugfix/race condition

pszujewski:
The fix in this PR is already in branch `chore/build-system`. I'm submitting this PR in case you are ok with bringing this bugfix into a future version 1 release instead of waiting for version 2. I'm not sure how far you are from a v2 release. 

The problem is a race condition created in index.ts. There is code that dispatches fetch requests in a loop that fails to `await` on the resolution of all the fetch requests. This race condition affected the state of `Publication.positions`, which was creating problems in the modules that used or read from this `positions` array. The solution was to use `Promise.all` to `await` on an array of promises to ensure all have resolved before allowing code execution to continue. 

I also added a brief note in the readme on how to "Get Started" with the examples locally. Thank you!


aferditamuriqi (approved):
Looks great, and tested , works as expected. thank you @pszujewski

PR #223, opened 2021-07-07
fix/package

kristojorg:
This is a small PR that tells webpack to build a UMD version of the package, which means it will be accessible for <script> tags and also for importing through CommonJS semantics.


aferditamuriqi (approved):
👍

PR #238, opened 2021-07-29
position calculation in sequence 

aferditamuriqi:
I've had to change the calculation of positions again, promise all and map seems to not generate the positions in sequence, but rather randomly.

first image shows out of sequence:
Screen Shot 2021-07-29 at 6 22 41 PM
second image schows with this fix:
Screen Shot 2021-07-29 at 6 20 43 PM

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -617,41 +617,41 @@ export async function load(config: ReaderConfig): Promise<any> {
     let positions = [];
     let weight = {};
     for (const link of publication.readingOrder) {
-        if ((publication.Metadata.Rendition?.Layout ?? "unknown") === "fixed") {
+      if ((publication.Metadata.Rendition?.Layout ?? "unknown") === "fixed") {
+        const locator: Locator = {
+          href: link.Href,
+          locations: {
+            progression: 0,
+            position: startPosition + 1,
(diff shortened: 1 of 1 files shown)


aferditamuriqi:
view file diff without whitespace https://github.com/d-i-t-a/R2D2BC/pull/238/files?diff=unified&w=1

kristojorg (approved):
> view file diff without whitespace https://github.com/d-i-t-a/R2D2BC/pull/238/files?diff=unified&w=1

Curious how you did this? This diff is way easier to read than the default one! 

In regards to the changed, this looks good to me. It shouldn't regress #222 because it does await each calculation before performing the next by using the `for` loop instead of the `map`.

pszujewski (approved):
This looks great. Yeah I'm looking closer at this now, and I see that the use of the variables totalContentLength and startPosition mean that the logic in the async map callback needed to be executed in series. Sorry I missed that before!

aferditamuriqi:
thank you both @kristojorg and @pszujewski 

@pszujewski no worries, i actually only figured that out while i was implementing sample read and wanted to make sure only a certain % is made available. and saw random chapters

PR #239, opened 2021-07-29
Feature: Sample-Read

aferditamuriqi:
This Feature allows any publication to be restricted as Sample or Preview.

how to use it:

in the reader config add the following section:

```
sample: {
     isSampleRead: true,
     limit: 10, // in percent e.g. 10%
     minimum: 5, // in pages
     popup: "You can borrow the book to continue reading." // any text that should be shown as overlay when the limit is reached
},
```

* minimum:
the minimum is the fallback for FXL publications, if the limit (in %) results in less than the minimum pages, for example:
10% of a FXL with 27 pages is 2 pages, that's not a lot, so if the minimum is greater than the pages resulting through the limit, than the minimum is used. 

Sample read is not a module, it's just a configuration to define the publication that is being rendered as sample read and limit reading beyond the set limits or minimum.

Please note: SampleRead relies on positions, so if you have disabled positions, SampleRead won't be able to calculate and compare based on percentage. 

you could have generation positions disabled:
```
rights: {
     autoGeneratePositions: false,
}
```
but if you provide services where the positions have been preprocessed, it would still work. as long as positions are available.
```
services: {
     positions: new URL("https://.../positions.json"),
},

```

### DIFF
--- a/src/modules/sampleread/SampleReadEventHandler.ts
+++ b/src/modules/sampleread/SampleReadEventHandler.ts
@@ -57,9 +57,6 @@ export default class SampleReadEventHandler {
     }
 
     function preventDefaultForScrollKeys(e) {
-      console.log(e.keyCode);
-      console.log(keys[e.keyCode]);
-      console.log(valid);
       if (keys[e.keyCode] && !valid) {
         preventDefault(e);
       }
@@ -133,7 +130,6 @@ export default class SampleReadEventHandler {
     }
 
     // IE9, Chrome, Safari, Opera
-    console.log("add sample event handlers");
     window.addEventListener("mousewheel", MouseWheelHandler, wheelOpt);
     // Firefox
     window.addEventListener("DOMMouseScroll", MouseWheelHandler, wheelOpt);
--- a/src/modules/sampleread/SampleReadEventHandler.ts
+++ b/src/modules/sampleread/SampleReadEventHandler.ts
@@ -1,0 +1,160 @@
+/*
+ * Copyright 2018-2021 DITA (AM Consulting LLC)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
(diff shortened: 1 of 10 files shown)


src/navigator/IFrameNavigator.ts
```diff
@@ -846,12 +854,12 @@ export default class IFrameNavigator implements Navigator {
     addEventListenerOptional(
       this.previousChapterAnchorElement,
       "click",
-      this.handlePreviousChapterClick.bind(this)
+      this.handlePreviousPageClick.bind(this)
```
pszujewski: Just curious, why are you making this change here? And also the similar one below? I would think that if you had an event listener on the `previousChapterAnchorElement`, then you would want the callback to be `handlePreviousChapterClick` instead of `handlePreviousPageClick`.

aferditamuriqi: Yes you are right @pszujewski, i am trying to solve an issue where the arrow left and right in the material menu which are `previousChapterAnchorElement` and `nextChapterAnchorElement` work on mobile devices, as there is not much room to have the buttons in the middle, but maybe i'll add new buttons for mobile. i'll give it a try and see if i can come up with something that doesn't lead to confusion.

thanks for pointing that out :) 

Screen Shot 2021-08-12 at 6 59 07 PM

src/modules/sampleread/SampleReadEventHandler.ts
```diff
@@ -0,0 +1,160 @@
...
+      return false;
+    }
+    function TouchStartHandler(e) {
+      e = e || window.event;
+      lastY = e.touches[0].clientY;
+      return false;
+    }
+
+    // IE9, Chrome, Safari, Opera
+    console.log("add sample event handlers");
```
kristojorg: Stray console log

src/modules/sampleread/SampleReadEventHandler.ts
```diff
@@ -0,0 +1,160 @@
...
+    };
+
+    function preventDefault(e) {
+      e.preventDefault();
+    }
+
+    function preventDefaultForScrollKeys(e) {
+      console.log(e.keyCode);
+      console.log(keys[e.keyCode]);
+      console.log(valid);
```
kristojorg: Stray console logs

kristojorg (dismissed):
I haven't tested this, but the code looks okay to me besides the stray console logs

src/modules/sampleread/SampleReadEventHandler.ts
aferditamuriqi: thank you :)

pszujewski (dismissed):
I tested out the branch locally and it seemed to work as expected. I tried with both "sample" on and off.

src/navigator/IFrameNavigator.ts
aferditamuriqi: reverted and changed the buttons to use api calls instead

PR #246, opened 2021-08-04
fixes an issue where navigating via TOC threw an exception.

aferditamuriqi:
remove anchor from href when finding positions for a resource. :)

### DIFF
--- a/src/model/Publication.ts
+++ b/src/model/Publication.ts
@@ -164,9 +164,8 @@ export class Publication extends R2Publication {
    * positionsByHref
    */
   public positionsByHref(href: string) {
-    return this.positions
-      ? this.positions.filter((el: Locator) => el.href === decodeURI(href))
-      : undefined;
+    const decodedHref = decodeURI(href) ?? "";
+    return this.positions.filter((p: Locator) => decodedHref.includes(p.href));
   }
 
   get hasMediaOverlays(): boolean {
--- a/src/model/Publication.ts
+++ b/src/model/Publication.ts
@@ -84,7 +84,7 @@ export class Publication extends R2Publication {
   }
   public getRelativeHref(href: string): string | null {
     const manifest = this.manifestUrl.href.replace("/manifest.json", ""); //new URL(this.manifestUrl.href, this.manifestUrl.href).href;
-    var href = href.replace(manifest, "");
+    var href = href.replace(manifest, "").slice(0, href.indexOf("#"));
     if (href.charAt(0) === "/") {
       href = href.substring(1);
     }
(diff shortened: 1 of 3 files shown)


kristojorg (approved):
A comment would be useful here, but it looks ok to me

pszujewski (dismissed):

kristojorg (changes requested):
So while this does stop it from throwing an error, there actually seems to be a larger issue here. In some cases, the location we want to go to is part of the way down a specific resource, like if one resource has two chapters in it. In that case, the url includes a `#` to reference the specific location. This is an extract of the `readingOrder` in an EPUB 2:
```
    {
      "title": "CHAPTER 1. Loomings.",
      "href": "OEBPS/@public@vhost@g@gutenberg@html@files@2701@2701-h@2701-h-1.htm.html#pgepubid00006"
    },
    {
      "title": "CHAPTER 2. The Carpet-Bag.",
      "href": "OEBPS/@public@vhost@g@gutenberg@html@files@2701@2701-h@2701-h-1.htm.html#pgepubid00007"
    },
```

So instead of just removing it, it looks like we need to use it to navigate. I just don't know how to interpret the `pgepubid00007` and turn that into a location. Any ideas?

aferditamuriqi:
@kristojorg i'll take a closer look, i did test it, and it seemed to navigate to the right place though, but i could be wrong.

aferditamuriqi:
moving this PR into draft, noticed a few more issues, the toc seems to work, but some other navigation aspects did break.

aferditamuriqi:
@kristojorg could you give this a test? with the book call of the wild, it seems all to work now (toc selection, next/previous chapter, etc.) i could be missing something.

kristojorg:
I think the issue comes only in publications that have two chapters on the same resource, so it's the inter resource navigation that breaks (well I think it was always broken maybe). I will try to find a publication with that setup, but it'll have to wait until Monday probably for me

kristojorg:
We can merge this because it doesn't break anything more than it was before, it's just that I noticed the problem after this PR.

aferditamuriqi:
The anchor navigation sure did work before, and should now as well, since this part is really only for finding the position not to navigate to, i'll find a book that has a TOC with anchors and i'll share it.

aferditamuriqi:
this books has sub chapters within the same resource, and all works as expected. btw this has been working before, the problem in this PR was that the positions where not found for the resource, as positions are generated for reading order and we are looking up positions by toc element, in that case we have to remove the fragment , but this has no impact on navigation. :) 
https://standardebooks.org/ebooks/william-congreve/the-way-of-the-world

here is a recording of this book https://www.dropbox.com/s/i7jhvuo148ld0bw/Screen%20Recording%202021-08-11%20at%2010.18.10%20AM.mov?dl=0

pszujewski (approved):
I pulled this branch down and ran it locally. Everything seemed to work as expected.

PR #247, opened 2021-08-11
Minor fixes.

aferditamuriqi:
- removed duplicate markup on dita example index, possibly caused  through merge conflicts
- removed streamed script
- applied some minor fixes reported through github actions.

### DIFF
--- a/src/modules/search/SearchModule.ts
+++ b/src/modules/search/SearchModule.ts
@@ -786,16 +786,16 @@ export default class SearchModule implements ReaderModule {
   }
 
   paginate(items: Array<any>, page: number, per_page: number) {
-    var page = page || 1,
-      per_page = per_page || 10,
-      offset = (page - 1) * per_page,
-      paginatedItems = items.slice(offset).slice(0, per_page),
-      total_pages = Math.ceil(items.length / per_page);
+    let _page = page || 1,
+      _per_page = per_page || 10,
+      offset = (_page - 1) * _per_page,
+      paginatedItems = items.slice(offset).slice(0, _per_page),
+      total_pages = Math.ceil(items.length / _per_page);
     return {
-      page: page,
-      per_page: per_page,
-      pre_page: page - 1 ? page - 1 : null,
-      next_page: total_pages > page ? page + 1 : null,
+      page: _page,
+      per_page: _per_page,
+      pre_page: _page - 1 ? _page - 1 : null,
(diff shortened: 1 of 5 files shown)


src/modules/search/SearchModule.ts
```diff
@@ -786,16 +786,16 @@ export default class SearchModule implements ReaderModule {
...
-    var page = page || 1,
-      per_page = per_page || 10,
-      offset = (page - 1) * per_page,
-      paginatedItems = items.slice(offset).slice(0, per_page),
-      total_pages = Math.ceil(items.length / per_page);
+    let _page = page || 1,
+      _per_page = per_page || 10,
+      offset = (_page - 1) * _per_page,
+      paginatedItems = items.slice(offset).slice(0, _per_page),
+      total_pages = Math.ceil(items.length / _per_page);
```
kristojorg, line 793: Just curious why you added the underscore in the name? I usually only do that for variables that will be unused

src/navigator/IFrameNavigator.ts
```diff
@@ -426,7 +426,7 @@ export default class IFrameNavigator implements Navigator {
     this.headerMenu = headerMenu;
     this.mainElement = mainElement;
     try {
-      var wrapper = HTMLUtilities.findRequiredElement(
+      let wrapper = HTMLUtilities.findRequiredElement(
```
kristojorg, line 429: Should this be a `const` if we are not reassigning it?

kristojorg (changes requested):
Couple small notes

src/modules/search/SearchModule.ts
aferditamuriqi: I didn't come up with a different name :) this can be anything really . But the actions where marking how it was before as being reassigned . And not valid .

src/navigator/IFrameNavigator.ts
aferditamuriqi: I guess it could , I am sure there are plenty of other places where that needs to be evaluated again too. This one was just a result of removing the re-declaration of it.

kristojorg: Ok yep we can address that later

src/modules/search/SearchModule.ts
kristojorg: I see yea I didn't notice that there was `_per_page` and `per_page` actually. Makes more sense now

Issue #229, opened 2021-07-11
V2 height issue 


PekopT:
https://brilliantov-screens.s3.eu-central-1.amazonaws.com/screencast_2021-07-12_00-27-45.mp4

something wrong with container height

PekopT:
@aferditamuriqi

aferditamuriqi:
@PekopT could you share you markup ? V2 is not entirely stable yet, especially not this container changes. But if you can share more info, we could look into it.

aferditamuriqi:
Fixed in https://github.com/d-i-t-a/R2D2BC/pull/248

PR #248, opened 2021-08-11
Several fixes related to the wrapper

aferditamuriqi:
- fixes position of selection toolbox in scroll mode, it was being added to the bottom instead of showing above the selection
- fixes height change on resize for fixed layout books, it was decreasing the height with each navigation. 
- some minor fixes on examples

fixes https://github.com/d-i-t-a/R2D2BC/issues/229


kristojorg (approved):
Code looks okay here as well

PR #249, opened 2021-08-12
matched implementation of  MediaOverlayIncrementable 

aferditamuriqi:
saw that the incremental of user-settings and TTS settings have this type, so i thought' i'd match the media overlay incremental as well.

### DIFF
--- a/src/reader.ts
+++ b/src/reader.ts
@@ -27,7 +27,7 @@ import MediaOverlayModule from "./modules/mediaoverlays/MediaOverlayModule";
 import {
   MediaOverlaySettings,
   IMediaOverlayUserSettings,
-  MediaOverlaySIncrementable,
+  MediaOverlayIncrementable,
 } from "./modules/mediaoverlays/MediaOverlaySettings";
 import TimelineModule from "./modules/positions/TimelineModule";
 import ContentProtectionModule from "./modules/protection/ContentProtectionModule";
@@ -406,7 +406,7 @@ export default class D2Reader {
     incremental:
       | UserSettingsIncrementable
       | TTSIncrementable
-      | MediaOverlaySIncrementable
+      | MediaOverlayIncrementable
   ): incremental is TTSIncrementable {
     return (
       incremental === "pitch" ||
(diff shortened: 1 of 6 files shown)


src/reader.ts
```diff
@@ -402,26 +403,48 @@ export default class D2Reader {
...
+  private isMOIncrementable(
+    incremental:
+      | UserSettingsIncrementable
+      | TTSIncrementable
+      | MediaOverlayIncrementable
+  ): incremental is MediaOverlayIncrementable {
+    return (
+      incremental === "pitch" ||
+      incremental === "rate" ||
+      incremental === "volume"
```
kristojorg: It looks like the `MediaOverlayIncrementable` is only for `"volume"` though right?

src/reader.ts
```diff
@@ -402,26 +403,48 @@ export default class D2Reader {
...
       if (this.navigator.rights?.enableTTS) {
         await this.ttsSettings.increase(incremental);
       }
+    } else if (this.isMOIncrementable(incremental)) {
+      if (this.navigator.rights?.enableMediaOverlays) {
+        await this.mediaOverlaySettings.increase(incremental);
+      }
     } else {
       await this.settings.increase(incremental);
     }
```
kristojorg, line 446: If both `MediaOverlaySettings` and `TTSIncrementable` use `"volume"`, how does this function know whether you are trying to increase one or the other? Currently if you call `d2Reader.increase("volume")`, only the `TTS` volume is actually increased. Are both meant to be increased or should we have `"TTS_volume"` and `"MediaOverlay_volume"`?

kristojorg (changes requested):
I think there is a small bug, but let me know if I am wrong!

src/reader.ts
aferditamuriqi: Yes but there will be rate at some point as well.

src/reader.ts
aferditamuriqi: ohh that's a good point, i'll test and see what happens

src/reader.ts
kristojorg: I think we should keep it to just what is implemented for now otherwise we could get errors. What this function is saying is that the passed in value _definitely_ conforms to the `MediaOverlayIncrementable` type, but in this case we know that is not true because neither `"pitch"` nor `"rate"` conform to that type currently

aferditamuriqi: i realized i didn't remove the once that are not implemented, just removed them.

src/reader.ts
aferditamuriqi: okey, so here is what i'll do, i'll rename the `volume` to `mo_volume` in media overlay, since that hasn't yet been used, but changing the tts names will break current implementations.

aferditamuriqi:
@kristojorg i renamed the `volume` to `mo_volume` and removed the not implemented ones

kristojorg (approved):

PR #250, opened 2021-08-12
extra pixels in last page in paginated mode 

aferditamuriqi:
applied a workaround how to calculate the remaining right scroll width which doesn't result in extra pixels.


kristojorg (approved):
Looks ok to me

PR #251, opened 2021-08-12
a few minor cleanups

aferditamuriqi:
- removed toggleDisplay function as this is a left over and not useful anymore
- minimal increase of timeout after resize
- markup cleanup


kristojorg (approved):
looks okay to me! I haven't tested but code looks fine

PR #252, opened 2021-08-18
Fixes Pagination

aferditamuriqi:
Issues were mostly visible in Firefox, but also present in other browsers under certain circumstances like slow connection etc.

issue:
- pagination would appear as one page instead of the actual page size
- navigation to a bookmark would break

fixes:
- refactored iFrame load and navigate to prevent race conditions.
- removed timeouts where an async function could be awaited instead,
- moved reporting of resource ready at the end
- moved setting the display positions to the end as well.
- removed flex box for reflowable publication 

and several other optimization ...

diff without whitespaces https://github.com/d-i-t-a/R2D2BC/pull/252/files?diff=unified&w=1

### DIFF
--- a/src/navigator/IFrameNavigator.ts
+++ b/src/navigator/IFrameNavigator.ts
@@ -2935,8 +2935,6 @@ export default class IFrameNavigator implements Navigator {
       }
       if (isCurrentLoaded) {
         console.log("is currently loaded");
-        console.log(locator.href);
-        console.log(this.currentChapterLink.href);
         if (locator.href.indexOf("#") !== -1) {
           const elementId = locator.href.slice(locator.href.indexOf("#") + 1);
           locator.locations = {
@@ -3064,7 +3062,7 @@ export default class IFrameNavigator implements Navigator {
             this.chapterTitle.innerHTML = "(Current Chapter)";
         }
 
-        await this.savePosition();
+        await this.updatePositionInfo();
       } else {
         if (this.searchModule !== undefined) {
           this.searchModule.clearSearch();
--- a/src/navigator/IFrameNavigator.ts
+++ b/src/navigator/IFrameNavigator.ts
@@ -2471,9 +2471,9 @@ export default class IFrameNavigator implements Navigator {
     this.handleResize();
   }
 
-  private async handlePreviousPageClick(
+  private handlePreviousPageClick(
     event: MouseEvent | TouchEvent | KeyboardEvent
-  ): Promise<void> {
+  ): void {
     this.stopReadAloud(true);
     if (this.view.layout === "fixed") {
       this.handlePreviousChapterClick(event);
@@ -2482,8 +2482,8 @@ export default class IFrameNavigator implements Navigator {
         this.handlePreviousChapterClick(event);
       } else {
         this.view.goToPreviousPage();
-        await this.updatePositionInfo();
-        await this.savePosition();
+        this.updatePositionInfo();
+        this.savePosition();
       }
       if (event) {
         event.preventDefault();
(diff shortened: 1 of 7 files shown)


aferditamuriqi:
I'll remove some of the console.logs once we are sure that all this is good to go.

src/navigator/IFrameNavigator.ts
```diff
@@ -2905,7 +2897,9 @@ export default class IFrameNavigator implements Navigator {
     }
   }
 
-  navigate(locator: Locator): void {
+  async navigate(locator: Locator): Promise<void> {
+    console.log("################")
```
pszujewski: Consider removing these two console logs?

src/navigator/IFrameNavigator.ts
```diff
@@ -1536,87 +1517,96 @@ export default class IFrameNavigator implements Navigator {
...
+        } else if (bookViewPosition > 0) {
+          this.view.goToPosition(bookViewPosition);
         }
-      }, 100);
+
+        this.newPosition = null;
+
+        this.hideLoadingMessage();
+        this.showIframeContents();
+        this.updatePositionInfo();
```
pszujewski: Would it make sense to call `this.updatePositionInfo();` before hideLoadingMessage and showIframeContents? Or it won't matter?

Hmm, I just realized this probably won't matter since hideLoadingMessage and showIframeContents doesn't impact the "Position Info" floating on the bottom

src/navigator/IFrameNavigator.ts
```diff
@@ -1423,31 +1425,10 @@ export default class IFrameNavigator implements Navigator {
       this.handleResize();
       this.updateBookView();
 
-      this.settings.applyProperties();
+      await this.settings.applyProperties();
```
pszujewski, line 1426: Does it make sense to put this before handleREsize and updateBookView still?

```
await this.settings.applyProperties();
this.handleResize();
this.updateBookView();
``

File: src/navigator/IFrameNavigator.ts [hunk 4]
```diff
@@ -2939,8 +2935,6 @@ export default class IFrameNavigator implements Navigator {
       }
       if (isCurrentLoaded) {
         console.log("is currently loaded");
```
pszujewski: Should this console log be wrapped if `if (IS_DEV) {}` or removed?

src/navigator/IFrameNavigator.ts
```diff
@@ -3228,6 +3215,18 @@ export default class IFrameNavigator implements Navigator {
         this.loadingMessage.style.display = "none";
         this.loadingMessage.classList.remove("is-loading");
       }
+      if (this.view.layout !== "fixed") {
+        console.log("hideLoadingMessage");
+        // this.checkResourcePosition();
```
pszujewski: Remove commented out code and console log?

pszujewski (dismissed):
Looks great! I left a few questions/comments/suggestions. I tested the changes locally on chrome and firefox. I throttled my network requests on firefox and didn't hit the bug. Everything else worked as expected in my testing, including bookmarks and highlights.

src/navigator/IFrameNavigator.ts
aferditamuriqi: removed

src/navigator/IFrameNavigator.ts
aferditamuriqi: removed

src/navigator/IFrameNavigator.ts
aferditamuriqi: removed

src/navigator/IFrameNavigator.ts
aferditamuriqi: i think this is fine

pszujewski (dismissed):

src/navigator/IFrameNavigator.ts
aferditamuriqi: i wanted to play it save and not move this yet.

PR #259, opened 2021-08-25
Bug/8101 - move initialization of content protection module to after reading location is updated

jdempcy:


jdempcy:
This pull request was opened by accident! Sorry about that!

PR #260, opened 2021-08-27
bugfix - Ensure keyboard event listener on window.document is only attached once

pszujewski:
`handleIframeLoad` is currently attaching the "keydown" event to the parent `window.document` every time the iframe load event fires. So, if you load new content into the iframe multiple times, pressing arrow left or right will skip a corresponding number of pages (in pagination mode).

The fix was to ensure that the event is only attached to the document element once during the first iframe "load" event. Thank you.


aferditamuriqi (dismissed):
good catch

aferditamuriqi (approved):

PR #261, opened 2021-08-30
Bug/8364

jdempcy:
Expand width of descramble area. I verified by trying to print that the descramble still works - it's still descrambled, I just no longer see descrambling artifacts when paging quickly, or scrambled text in the top-left paragraph when that paragraph originated on the previous screen.


aferditamuriqi (changes requested):
@jdempcy it seems like a lot of unrelated commits, i am having some difficulty to review this PR.

jdempcy:
it was a mistaken pull request to the upstream repo instead of the downstream fork, sorry about that!

PR #276, opened 2021-10-08
Add setIframeHeight method to paginated state

Toxiapo:
This PR fixes the iframe height issue when switching scroll mode to paginated mode from the initial load.

An issue I was having while using the v2-branch was that when I reload the page on scroll mode and switch to paginated mode, the iframe height was 0, causing a blank page. 

The following proposed fix is to add the `this.setIframeHeight(this.iframe);` method to [setMode](https://github.com/d-i-t-a/R2D2BC/blob/1c11193d83d75957c97f8ae921a14564073b027b/src/views/ReflowableBookView.ts#L49).

### DIFF
--- a/src/views/ReflowableBookView.ts
+++ b/src/views/ReflowableBookView.ts
@@ -97,6 +97,7 @@ export default class ReflowableBookView implements BookView {
       ) as any;
       html.style.setProperty("--USER__scroll", "readium-scroll-off");
       this.setSize();
+      this.setIframeHeight(this.iframe);
     }
     if (this.delegate.rights?.enableContentProtection) {
       this.delegate.contentProtectionModule.recalculate();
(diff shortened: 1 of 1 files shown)


aferditamuriqi (commented):
Hey @Toxiapo , i am reviewing your PRs, and it looks like this does fix it only temporary, what happens is that this code get's actually called in the wrong mode. meaning even-though it should be in paginated , it's actually still in scroll-mode. i have had a few issues about the iframe height and i have solved all of them in a recent feature branch. what i would like to do is, merge up those fixes first and bring v2 up to date as well, and ask you to please test those again. i will let you know once i have it all merged.

aferditamuriqi:
@Toxiapo i've merged the Annotations work and the resize fixes into V2, please give it a try if this issue still persists, i believe it isn't there anymore. i will be publishing a beta version of v2 as well.

Toxiapo:
Thanks, @aferditamuriqi! I will try out the new merge.

aferditamuriqi (approved):

PR #277, opened 2021-10-13
Bugfix: internal links in remote-hosted epubs open in new tab

hokei:
This PR fixes an issue where, when an epub is hosted on a separate remote server, its internal links are considered external and open in a new tab.  

You can see an example of this issue here:  https://drb-qa.nypl.org/read/6585394 (navigate to "contents") 

This doesn't seem to be a complete fix - this fix will not catch the case where the same host serves both the epub and also some other content - eg:  if `www.company.com` serves `wwww.company.com/files/epub_id/manifest.json` and this epub links to `www.company.com/about/team`, this second page will be interpreted as internal to the epub.  

I'd like to know how much of an edge case this is - I think if it's possible to check the link location against the manfiest URL, that might give us more confidence, but I'm not quite sure how to get the manifest URL inside the `EventHandler`, or if we even want to.


aferditamuriqi (commented):
Hello @hokei , indeed this is a bit tricky, but i would like to implement it based on your suggestion with the manifest. even though your changes would work, they would also break books that have links that should be external but come from the same referrer. if ok with you i can help you get this in the right direction.

hokei:
@aferditamuriqi This sounds good!  Yes, if you have ideas on how the `EventHandler` can access the `manifestUrl` that'd be super helpful to me!  Right now it's all being called via a listener, so I don't know how this information could be passed.

aferditamuriqi:
> @aferditamuriqi This sounds good! Yes, if you have ideas on how the `EventHandler` can access the `manifestUrl` that'd be super helpful to me! Right now it's all being called via a listener, so I don't know how this information could be passed.

@hokei yes, in some WIP i am actually adding the navigator as argument into the event handler, i can pull in those changes easily upfront, that way you would have access to the entire publication, and therefor to the manifest as well. In my work i am refactoring the footnotes, for which i also need access to the publication in the event handler

src/utils/EventHandler.ts
```diff
@@ -80,15 +86,36 @@ export default class EventHandler {
...
         event.stopPropagation();
       } else {
         (event.target as HTMLAnchorElement).href = link.href;
         if (isSameOrigin && isInternal !== -1) {
           this.onInternalLink(event);
+          const link = event.target as HTMLLIElement;
+          if (link) {
+            const attribute = link.getAttribute("epub:type") === "noteref";
+            if (attribute) {
+              // await this.popup.handleFootnote(link, event);
```
hokei, line 112: This can be uncommented when the footnote work is merged.  I can add a TODO here if needed, or remove the whole block, depending on the order of things getting merged.

aferditamuriqi: 👍

aferditamuriqi (approved):

PR #278, opened 2021-10-13
Bugfix: next page button becomes inoperable on scrollMode for some books

Toxiapo:
This PR updates the atEnd function to use `clientHeight` instead of `scrollHeight` for calculating the remaining page content.

The bug appears when loading a reflowed page on scroll mode on firefox. On scroll mode, some margin and padding were added through CSS to the  <body> tag inside the Iframe document, making the Iframe to overflow, clientHeight returns the actual viewable height of the document, which is what we want.

## To Reproduce

- run `npm run example:react` with this book. https://drb-files-qa.s3.amazonaws.com/epubs/gutenberg/65303_noimages/manifest.json
- Switch to scroll mode, the next page button does not take you to the next resource.

### DIFF
--- a/examples/react/index.tsx
+++ b/examples/react/index.tsx
@@ -62,6 +62,7 @@ const App = () => {
             id="iframe-wrapper"
             style={{
               height: "calc(100vh - 10px)",
+              overflow: "hidden",
             }}
           >
             <div id="reader-loading" className="loading"></div>
--- a/src/views/ReflowableBookView.ts
+++ b/src/views/ReflowableBookView.ts
@@ -325,7 +325,7 @@ export default class ReflowableBookView implements BookView {
         "#iframe-wrapper"
       ) as HTMLDivElement;
       return (
-        Math.ceil(this.scrollingElement.scrollHeight - wrapper.scrollTop) - 1 <=
+        Math.ceil(this.scrollingElement.clientHeight - wrapper.scrollTop) - 1 <=
         BrowserUtilities.getHeight()
       );
     } else {
--- a/src/views/ReflowableBookView.ts
+++ b/src/views/ReflowableBookView.ts
@@ -84,8 +84,6 @@ export default class ReflowableBookView implements BookView {
       this.setSize();
       this.setIframeHeight(this.iframe);
     } else {
-      this.height =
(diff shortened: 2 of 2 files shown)


hokei:
Hmmm this seems like a weird bug - I was able to reproduce the original non-working version only on Firefox, and confirm that it is fixed in the new version.

aferditamuriqi (commented):
@Toxiapo let me test this out first, because client height is actually not really what i wanted to use here, the atEnd should give back when you actually are at the end of a resource. next page buttons are inoperable if the resource reaches the end, this is by design. but maybe there is something else that you need for this. or i am misunderstanding the issue :) either-way, it might good if you joined us in slack .

Toxiapo:
> @Toxiapo let me test this out first, because client height is actually not really what i wanted to use here, the atEnd should give back when you actually are at the end of a resource. next page buttons are inoperable if the resource reaches the end, this is by design. but maybe there is something else that you need for this. or i am misunderstanding the issue :) either-way, it might good if you joined us in slack .

@aferditamuriqi My expectation was that atEnd should return true if I am at the bottom of the page, but it does not sometimes.

But yea I think slack chat would be easier for further discussion!

aferditamuriqi:
> > @Toxiapo let me test this out first, because client height is actually not really what i wanted to use here, the atEnd should give back when you actually are at the end of a resource. next page buttons are inoperable if the resource reaches the end, this is by design. but maybe there is something else that you need for this. or i am misunderstanding the issue :) either-way, it might good if you joined us in slack .
> 
> @aferditamuriqi My expectation was that atEnd should return true if I am at the bottom of the page, but it does not sometimes.
> 
> But yea I think slack chat would be easier for further discussion!

Yes it sure should return true if in scrollmode you reach the bottom of the iframe content. but when using clientHeight it would return always true, even though it's not at the end. since the clientHeight is only the visible area.

Toxiapo:
@aferditamuriqi Thanks for taking a look at this PR! scrollHeight seems to be the correct API used here, I will revert that change and will be keeping the overflow setting on the example page because that seems to be what we need to get firefox to work.

aferditamuriqi (approved):

PR #296, opened 2021-11-09
bugfix - Removes use of debugger checker

pszujewski:
This PR updates the devtools detector library in order to remove use of the lib's "debugger checker", which forced only Chrome to "pause" on a debugger statement every time you opened in the devtools in content protection/detect inspect mode. The point of allowing the user to detect inspection is to protect against someone maliciously accessing the JS console, or inspecting the state of the DOM. This arbitrary debugger "pause" in chrome made it possible to still make use of the browser's dev tools. Luckily, this behavior can be removed simply by explicitly opting out of the debugger checker in the code, as I do in this PR.

### DIFF
--- a/src/modules/protection/ContentProtectionModule.ts
+++ b/src/modules/protection/ContentProtectionModule.ts
@@ -27,7 +27,7 @@ import {
 import { debounce } from "debounce";
 import { IS_DEV } from "../../utils";
 import { delay } from "../../utils";
-import { addListener, launch } from "devtools-detector";
+import { DevtoolsDetector, checkers } from "devtools-detector";
 import { getUserAgentRegExp } from "browserslist-useragent-regexp";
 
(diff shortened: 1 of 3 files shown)


src/modules/protection/ContentProtectionModule.ts
```diff
@@ -740,8 +749,11 @@ export default class ContentProtectionModule implements ReaderModule {
     if (IS_DEV) {
       console.log("before print");
     }
-    this.delegate.headerMenu.style.display = "none";
-    this.delegate.mainElement.style.display = "none";
+
+    if (this.delegate && this.delegate.headerMenu) {
```
pszujewski, line 753: I added this conditional because I was seeing this.delegate is undefined errors. It seemed harmless to add this check.

aferditamuriqi (approved):

PR #301, opened 2021-11-23
Feature: FXL annotations

aferditamuriqi:


aferditamuriqi:
Tagging as a v3 candidate. FXL annotations will be revisited as part of the v3 architecture overhaul.

PR #306, opened 2021-12-08
V2 Review and Testing

aferditamuriqi:
@kristojorg I am finalizing my testing, reviewing and fixing issues where they come up, as well as adding documentation. over the next ~2 weeks i will be adding to the draft PR and then once finalized and you get a change to review, i am ready to merge up to develop. targeting to finalize it before the holidays :) 

@pszujewski @Toxiapo @hokei please feel free to share your feedback as well. :) 

as usual i am combining commits based on feature/fixes/cleanup etc. for easier review. 

included in this PR also this https://d-i-t-a.github.io/R2D2BC/classes/D2Reader.html (keep in mind this is WIP as i keep adding comments within the code) 

fixes https://github.com/d-i-t-a/R2D2BC/issues/294 (isScrollMode sometimes is called too early)
fixes https://github.com/d-i-t-a/R2D2BC/issues/293 (Paginated view text truncation in v2)
fixes https://github.com/d-i-t-a/R2D2BC/issues/215 (Make findRequiredElement generic V2)
fixes https://github.com/d-i-t-a/R2D2BC/issues/214 where possible (Make methods into getters wherever possible on D2Reader V2)
fixes https://github.com/d-i-t-a/R2D2BC/issues/206 (Enable strictNullChecks V2)
fixes https://github.com/d-i-t-a/R2D2BC/issues/205 for rights (Distinct module input options from internal settings V2)

PR #315, opened 2022-01-21
Reset iframe height on each load

Toxiapo:
I am experiencing a bug on some contents that, when a user navigates to a page with a large resource and navigates back, the iframe keeps the largest height value for the iframe, which results in a page with excessive blank areas.

### To Reproduce
1. Replace the [example book](https://github.com/d-i-t-a/R2D2BC/blob/1e374e820fdf625e9f3918abc085e414c951fb70/examples/react/index.tsx#L19) with this. https://drb-files-qa.s3.amazonaws.com/epubs/gutenberg/65303_noimages/manifest.json
2. run `npm run example:react`, and open the page on Chrome or Safari
3. Turn on 'scrolling mode', and get to page 2
4. Now if you press the back button, and get to page 1, you will see a huge page with a large blank area under the cover page and a big vertical scroll bar.

I assume it was because when the iframe loads, it somehow saves the old iframe height value or a race condition somewhere that made the 'setIframeHeight` function references the previous iframe instance.

This PR resets the iframe height before it loads the content, and it seems to be working fine on chrome, safari, or firefox. This seems trivial and I am not sure if this is the right place to set this, please let me know what you think!

### DIFF
--- a/src/navigator/IFrameNavigator.ts
+++ b/src/navigator/IFrameNavigator.ts
@@ -1380,6 +1380,10 @@ export default class IFrameNavigator implements Navigator {
       if (this.newPosition) {
         bookViewPosition = this.newPosition.locations.progression;
       }
+
+      // Reset before setIframeHeight is called to calculate the height.
+      this.iframes[0].height = "0px";
+
       await this.handleResize();
       this.updateBookView();
 


aferditamuriqi:
Peter and I have been literally working on this . Seems like an issue that only is reproducible in a react environment, but I am close to having a solution.. I'll review your change here as well but I'd rather not make this change here. I'll keep you posted .

aferditamuriqi:
@Toxiapo i have added this fix in v1, which could be pulled into v2 as well, if you would like to test it out, https://github.com/d-i-t-a/R2D2BC/commit/f4cdd7c56ae74e9e41d6cea6741c94e29703c25b , i'd like to set the 0 size in the setIframeHeight function itself.

aferditamuriqi:
@Toxiapo could you try this PR https://github.com/d-i-t-a/R2D2BC/pull/306 , i have a bunch more scroll mode issues fixed there.

Toxiapo:
@aferditamuriqi Ahh ok, that makes sense, thank you! I will close this then and look at that PR.

PR #323, opened 2022-01-26
recalculate in handleIframeLoad

pszujewski:
It's necessary to call `this.contentProtectionModule.recalculate` if the active book position is advanced beyond the beginning of a chapter in `handleIframeLoad`. Otherwise, the text appears scrambled:

content_is_scrambled

### DIFF
--- a/src/modules/protection/ContentProtectionModule.ts
+++ b/src/modules/protection/ContentProtectionModule.ts
@@ -838,21 +838,28 @@ export default class ContentProtectionModule implements ReaderModule {
     }
   }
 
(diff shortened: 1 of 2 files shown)


aferditamuriqi:
@pszujewski makes sense, i'll test it, merge it and tag a new version by tomorrow.

aferditamuriqi (approved):
@pszujewski this looks good, i'll merge this here, but will need to cherry pick it out into the production branch to tag a  version 1.11.3 with this. develop has still a beta feature which i will be merging soon too, but not yet.

PR #356, opened 2022-04-29
Removed unused fetch request

aferditamuriqi:
the result of the fetch request was never used.


bgoldowsky (approved):
Other than a lot of === vs. == warnings from the linter (probably harmless?), this looks fine to me.

PR #358, opened 2022-05-10
settings for customKeyboardEvents

aferditamuriqi:
Added customKeyboardEvents setting under rights, the default is false which keeps the left/right arrow keys for paging forward / backward.

additionally cleaned up dead code


mbrambilla:
Seems reasonable to me.
Is something similar going to occur for line focus up/down arrows?

aferditamuriqi:
> Seems reasonable to me. Is something similar going to occur for line focus up/down arrows?

possible, but i didn't want to make changes there until we finalize the line focus feature.

mbrambilla:
All good then.

PR #366, opened 2022-07-14
Update moment's version

audieleon:
Addresses a vulnerability in moment.


aferditamuriqi:
Hi @audieleon i appreciate the PR, but the way the vulnerabilities are addressed, is in the package-lock, the package.json is already setup properly to use any update minor or patch on that version. see ^ sign. 
With each new version of the reader we address vulnerabilities with the npm audit fixes. 

see package lock here https://github.com/d-i-t-a/R2D2BC/blob/develop/package-lock.json#L21237

PR #412, opened 2022-12-13
Updated devtools detector version

jdempcy:
This fixes an issue where devtools detector fills the logs with a 500-item JSON object on every page load.


jdempcy:
Closing this because there is an issue with the whitespace. Showing 19,000 lines changed, which is obviously not correct.

PR #413, opened 2022-12-13
Updated devtools-detector to 2.0.14 

jdempcy:
Fix for 500-item JSON object being logged 25 times on every pageload, making it really hard to develop in Remix where the console logs are outputted to the server output.


jdempcy:
Once again it is showing 19,000 lines changed. I'll try using a different editor to make the change, as VS Code seems to be overriding the whitespace preferences. (Tabs vs. spaces.)

PR #427, opened 2023-01-30
Bugfix/reduce sentry errors

jdempcy:
This PR catches a number of errors that are showing up in Sentry and logs them instead of throwing them as errors.

Note: Please review, as this may in some cases hide larger issues or make it harder to debug down the road.

Catching these errors is only appropriate if they are known issues that do not have any user-facing side effects. In that case, it will help us greatly reduce the errors reported in Sentry so we may more easily debug other issues.


jdempcy:
New PR here: https://github.com/d-i-t-a/R2D2BC/pull/428

PR #434, opened 2023-03-06
Feature/improve annotations loading flow

niko-8d4:
When a user navigates to another chapter, both navigate and handleIFrameLoad are being called. There is some overlap between the two functions and e.g. the annotations are being drawn multiple times. Some of those happens too early.

This PR makes it so that navigating between chapters only redraws the annotations once and allows to specifically skip redrawing of annotations when the book view is updated (only relevant for handleIFrameLoad for now).

### DIFF
--- a/src/modules/AnnotationModule.ts
+++ b/src/modules/AnnotationModule.ts
@@ -492,9 +492,7 @@ export class AnnotationModule implements ReaderModule {
   public showHighlights() {
     let highlights: Array<any> = [];
     if (this.annotator) {
-      highlights = this.annotator.getAnnotationsByChapter(
-        this.delegate.currentLocator().href
-      ) as Array<any>;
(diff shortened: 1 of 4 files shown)


src/modules/AnnotationModule.ts
```diff
@@ -492,7 +492,9 @@ export class AnnotationModule implements ReaderModule {
   public showHighlights() {
```
aferditamuriqi: @nikolai-ri i missed this when we spoke, we have to revert the change on the highlights list in this method, as this method is solely responsible for showing all highlights in the toc area , see dita reader example.

niko-8d4: Thanks & good catch!
I reverted the change in showHighlights to get the annotations of all chapters and can confirm its working in the reader example project.

aferditamuriqi (approved):

PR #435, opened 2023-03-08
Analytics

KaramYaaqba:
add analytics module and open book event


aferditamuriqi (changes requested):
@KaramYaaqba we can talk about this on the call, but there is a lot of that is hardcoded in this PR and a lot that is incomplete, as well as naming and cleanup of the code has not been done.

aferditamuriqi (changes requested):
@KaramYaaqba based on our last conversation, we should close this PR

PR #458, opened 2023-04-10
move code that converts lrp href to fully-qualified to before comparison with lastInHistory href

bluefirepatrick:
The bug happens in the HistoryModule when the lastReadingPosition progression is something greater than 0 and the lastReadingPosition href is compared with the lastInHistory href.  In the bug scenario, two entries are added to the history array -which I believe is unintended.

There are two places in the HistoryModule where items can be added to the history array: line 151 (PUSH 1) and line 164 (PUSH 2). Both PUSHs are called when the progression of the LRP is greater than 0 and the href of the LRP is not equal to the lastInHistory href.  Unfortunately the LRP href is based on tocItem (which is a relative path from the manifest) while lastInHistory href is always fully-qualified because PUSH 1 converts it before pushing and PUSH 2 is based on the locator param passed in from IFrameNavigator. Because of this there are scenarios where both PUSH 1 and PUSH 2 can add entries to the history array when, I believe, only one or the other is intended to.   

NOTE: The bug occurs when the progression of the LRP is greater than 0 -so you have to scroll down slightly in scrolling mode or go to the next chapter page in paginated.   

The fix is to move the code that converts the LRP href value to fully-qualified to before the comparison. In my testing so far (in three browsers) this seems to work well.

### DIFF
--- a/src/modules/history/HistoryModule.ts
+++ b/src/modules/history/HistoryModule.ts
@@ -137,16 +137,16 @@ export class HistoryModule implements ReaderModule {
           this.history = this.history.slice(0, this.historyCurrentIndex);
         }
         lastInHistory = this.history[this.history.length - 1];
+        const linkHref = this.publication.getAbsoluteHref(
+          lastReadingPosition.href
+        );
+        log.log(lastReadingPosition.href);
+        log.log(linkHref);
+        lastReadingPosition.href = linkHref;
         if (
           (lastInHistory && lastInHistory.href !== lastReadingPosition.href) ||
           lastInHistory === undefined
         ) {
-          const linkHref = this.publication.getAbsoluteHref(
-            lastReadingPosition.href
-          );
-          log.log(lastReadingPosition.href);
-          log.log(linkHref);
-          lastReadingPosition.href = linkHref;
 
           this.history.push(lastReadingPosition);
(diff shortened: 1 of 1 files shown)


bluefirepatrick:
@aferditamuriqi this new commit fixes two issues: 'back button then new location' and 'return to proper scroll/page locations in history traversals' as well as the original 'comparing local and fully-qualified domains'.  I have sent you more info about these is slack.  In my testing so far things appear to be working well.  Happy to talk this over.

src/modules/history/HistoryModule.ts
```diff
@@ -102,21 +102,25 @@ export class HistoryModule implements ReaderModule {
           " disabled",
           ""
         );
+        this.historyForwardAnchorElement.hidden = false;
```
aferditamuriqi: Ok, everything else works as it's supposed to, but i would rather not hide the forward or backward buttons, as these are used in the example, for the implementation any UI can be used other than the sample UI

bluefirepatrick: Sure, just pushed a commit so that they are no longer being hidden.  It just made it easier to test.

aferditamuriqi (dismissed):

aferditamuriqi (approved):

PR #589, opened 2023-10-20
Improve performance of mouse move event handler

niko-8d4:
- Modifies the processMouseEvent function.
- This function runs on every mouse move, thus potentially slowing down the reader, if users have many highlights.
- The part that checks if we are hovering a highlight, is now moved to event listeners directly on the highlight dom elements. 
- The processMouseEvent function now only processes mouseup/click/touchup events

### DIFF
--- a/src/modules/highlight/TextHighlighter.ts
+++ b/src/modules/highlight/TextHighlighter.ts
@@ -362,9 +362,8 @@ export class TextHighlighter {
   }): Array<any> {
     let order: any[] = [],
       chunks: any = {},
-      grouped:
-        | any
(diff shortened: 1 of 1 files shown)


src/modules/highlight/TextHighlighter.ts
```diff
@@ -2948,7 +2895,7 @@ export class TextHighlighter {
         }
       }
 
-      highlightArea.style.setProperty("pointer-events", "none");
+      highlightArea.style.setProperty("pointer-events", "all");
```
aferditamuriqi, line 2906: this might be an issue as not all layers have pointers.

aferditamuriqi (approved):
will fix the pointer events in next PR

PR #661, opened 2023-11-15
get landmarks from the reader instance

### LINKED ISSUE #640: get Landmarks from the reader instance
Hi,

it is possible to get the table of contents from a reader instance with `d2Reader.tableOfContents`.

It would be nice to retrieve the list of landmarks with an equivalent call, like `d2Reader.landmarks`

gotson:
Closes:  #640


aferditamuriqi (approved):
looks good, thank you for your contribution @gotson

PR #662, opened 2023-11-15
configuration to use a memory store

### LINKED ISSUE #628: disable storage?
Hi,

there is an option to switch between local and session storage, but is it possible to opt-out entirely of settings storage ?

I want to manage settings storage myself, and i don't want the reader to store anything.

gotson:
Closes: #628

@aferditamuriqi it seems there is already a `Store` interface, and a `MemoryStore`.

The reader exposes a `ReaderConfig.useLocalStorage` configuration, which is passed to the `LocalStorageStore`, and defines whether the store uses `window.localStorage` or `window.sessionStorage` for storage.

This PR adds a new configuration key `ReaderConfig.storageType`. By default this is unset, and will not change the default behaviour, as it would use a `LocalStorageStore`.

If the key is set to `memory`, then a `MemoryStore` is used.

I used a `string` configuration key, so it can be expanded in the future if needed.

### DIFF
--- a/src/navigator/IFrameNavigator.ts
+++ b/src/navigator/IFrameNavigator.ts
@@ -212,6 +212,7 @@ export interface ReaderConfig {
   injectables: Array<Injectable>;
   injectablesFixed?: Array<Injectable>;
   useLocalStorage?: boolean;
+  storageType?: string;
   attributes?: IFrameAttributes;
   services?: PublicationServices;
   sample?: SampleRead;
--- a/src/reader.ts
+++ b/src/reader.ts
@@ -45,6 +45,8 @@ import {
 } from "./navigator/IFrameNavigator";
 import LocalAnnotator from "./store/LocalAnnotator";
 import LocalStorageStore from "./store/LocalStorageStore";
+import Store from "./store/Store";
+import MemoryStore from "./store/MemoryStore";
 import { findElement, findRequiredElement } from "./utils/HTMLUtilities";
 import { convertAndCamel } from "./model/Link";
 import { LayerSettings } from "./modules/highlight/LayerSettings";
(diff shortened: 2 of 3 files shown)


aferditamuriqi:
@gotson i will review and test it, if this works in other scenarios, i am not seeing how you would be able to use this for example, if you set font size, and switch to the next chapter, it would loose those settings. if that is the case, i won't be able to accept this PR, but open to updates, i will let you know here soon when i have reviewed and tested. thank you for your contribution. :)

gotson:
> Closes: #628
> 
> @aferditamuriqi it seems there is already a `Store` interface, and a `MemoryStore`.
> 
> The reader exposes a `ReaderConfig.useLocalStorage` configuration, which is passed to the `LocalStorageStore`, and defines whether the store uses `window.localStorage` or `window.sessionStorage` for storage.
> 
> This PR adds a new configuration key `ReaderConfig.storageType`. By default this is unset, and will not change the default behaviour, as it would use a `LocalStorageStore`.
> 
> If the key is set to `memory`, then a `MemoryStore` is used.
> 
> I used a `string` configuration key, so it can be expanded in the future if needed.

It doesn't. It loses all settings if you refresh the whole page, but when changing chapter the reader instance is not recreated, thus the memory settings are kept.

aferditamuriqi:
@gotson i think i would like to optimize this a little more. so that the new setting can actually be used instead of the useLocalStorage but also make sure it's backwards compatible. so that the new setting can take memory, local, and session and handle it properly when set

aferditamuriqi:
@gotson please see if this works for you as well https://github.com/d-i-t-a/R2D2BC/pull/758

aferditamuriqi:
Closing this PR in favor of already merged solution

PR #677, opened 2023-11-22
correctly store previous reading position in history

### LINKED ISSUE #676: History not kept when navigating with previous/next chapter buttons
I was testing the History feature and I was getting erratic behaviour, for instance:
- open book
- move forward using keyboard arrow right
- encounter a link to the bibliography, click on it
- click history back => at that point i expect to be brought back to the page i was before clicking the bibliography link, but instead i was brought back to the cover page

If however i navigate using the TOC to a chapter, then click a bibliography link, i am taken back to the chapter i clicked on the TOC.

~~It seems this is because of calls `this.navigate(position, false)` in the `IFrameNavigator`, which do not persist the history.~~

~~Removing the `, false)` solve the problem, however i would like to check with you, as this may be there for a reason i can't fathom.~~

gotson:
Closes: #676

### DIFF
--- a/src/modules/history/HistoryModule.ts
+++ b/src/modules/history/HistoryModule.ts
@@ -135,10 +135,7 @@ export class HistoryModule implements ReaderModule {
           this.history = this.history.slice(0, this.historyCurrentIndex);
           this.history.push(lastReadingPosition);
           this.historyCurrentIndex = this.history.length - 1;
-        } else if (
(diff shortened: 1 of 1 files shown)


aferditamuriqi:
this PR should be replaced by this one https://github.com/d-i-t-a/R2D2BC/pull/784

gotson:
> this PR should be replaced by this one #784

tested and seems good, closing

PR #679, opened 2023-11-23
prefix css classes, use ID selectors when possible

### LINKED ISSUE #641: css classes conflict with popular framework
Hi,

i noticed that the css classes used in this project are not prefixed, and may conflict with the ones used in popular frameworks like bootstrap or vuetify.

Example of these are `info` and `error`.

It's a good practice to prefix css classes so that they don't conflict.

gotson:
This PR addresses #641 by prefixing some CSS classes with `r2d2bc-`.

It also removes some of the logic to find HTML elements by CSS classes, and instead uses ID selectors.

Closes #641

### DIFF
--- a/examples/react/index.tsx
+++ b/examples/react/index.tsx
@@ -65,7 +65,7 @@ const App = () => {
             }}
           >
             <div id="reader-loading" className="loading"></div>
-            <div id="reader-error" className="error"></div>
+            <div id="reader-error"></div>
           </main>
         </div>
       </div>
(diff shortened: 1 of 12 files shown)


gotson:
i've added one more commit to fix some CSS conflicts coming from the TTS styles. Since the project does not use scoped styles, the stylesheets impact the whole application, and thus applying very generic styles like `input[type=checkbox]` is not a good practice.

gotson:
hi @aferditamuriqi do you think this PR could be looked at? thanks

aferditamuriqi:
Tagging as a v3 candidate. The CSS class prefixing is a breaking change for existing integrators, so it fits better in the next major release where we can bundle it with other breaking changes.

Issue #627, opened 2023-11-06
event listener for unhandled clicks/touch?


gotson:
Hi,

is it possible to add an event listener for clicks or touch events that were not handled by the reader?

I would like to display my own menu when a click/touch event happens, so if such event happen, and is not acted upon by the reader itself, my callback would be called.

aferditamuriqi:
you can do this by injecting your own script. see injectables.

gotson:
> you can do this by injecting your own script. see injectables.

i am not sure how to use that, would you be able to provide some guidance or basic very simple example ? I only need some kind of callback for unhandled clicks, i will handle the menu display and everything outside the reader.

aferditamuriqi:
@gotson there is a basic implementation of a click script in the injectables samples. look into the index_api.html and for the injectables section, there is for example an injectable like this

`            { type: 'script', url: '/viewer/injectables/click/click.js' },
`

gotson:
> @gotson there is a basic implementation of a click script in the injectables samples. look into the index_api.html and for the injectables section, there is for example an injectable like this
> 
> ` { type: 'script', url: '/viewer/injectables/click/click.js' },`

Thanks for pointing this out. It works, but it does not solve my problem. Event bubbling is limited to the current document, and cannot bubble up from an iFrame to the containing document.

The sample is fine to output something in the console, but i don't see how i can use that to trigger something in the document containing the reader.

PR #681, opened 2023-11-27
Feature: event fallthrough for keydown and click

gotson:
This PR addresses #627

As mentioned in [my last comment](https://github.com/d-i-t-a/R2D2BC/issues/627#issuecomment-1820555473) on the above issue, event bubbling from an iframe to the container does not work in Javascript, thus injectables do not solve the problem at hand.

Since the `IFrameNavigator` is handling the iframes themselves, and can be configured by the upper component, it makes sense that it allows for event redirection.

The `clickThrough` functions were already there, but not exposed. I added a new API field for this, and i had to change a few things to correctly handle the popup clicks, so that the events are not sent in the `clickThrough` API if they already had an effect in the iFrame. For instance when clicking to reduce the image/footnote popup, the click event will not be falling through.

I implemented something similar for keyboard events, where any keydown event that was not being handled in the Navigator will be falling through to the containing component.

IMHO this is an elegant solution, which allows for out of the box customization without having to resort to injectables.

### DIFF
--- a/src/modules/search/Popup.ts
+++ b/src/modules/search/Popup.ts
@@ -151,16 +151,21 @@ export class Popup {
           return;
         }
         let self = this;
-        win.onclick = function (ev) {
-          if (event.target !== ev.target) {
-            if (d2popover.parentElement) {
-              self.hidePopover();
-              if (win) {
-                win.onclick = null;
+        win.addEventListener(
+          "click",
+          function (ev) {
+            if (event.target !== ev.target) {
+              if (d2popover.parentElement) {
+                self.hidePopover();
+                ev.stopImmediatePropagation();
               }
             }
+          },
+          {
+            once: true,
+            capture: true,
           }
-        };
+        );
       }
     } else if (src) {
       let absolute = getAbsoluteHref(src);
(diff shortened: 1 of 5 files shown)


src/modules/search/Popup.ts
```diff
@@ -207,16 +212,21 @@ export class Popup {
           return;
         }
         let self = this;
-        win.onclick = function (ev) {
-          if (event.target !== ev.target) {
-            if (d2popover.parentElement) {
-              self.hidePopover();
-              if (win) {
-                win.onclick = null;
+        win.addEventListener(
```
aferditamuriqi, line 215: I am not seeing anywhere where this event listener is removed.

aferditamuriqi:
@gotson i will run this locally to better understand and test the PR

aferditamuriqi (approved):
LGTM

aferditamuriqi:
@gotson can you please resolve the conflict and also make sure linter is run , thank you , i'll merge it once you have an updated PR

aferditamuriqi (changes requested):
please resolve conflict

gotson:
will have a look when time allows

gotson:
@aferditamuriqi rebased and linted 👍🏻

aferditamuriqi (approved):

PR #697, opened 2023-11-28
fix: data-viewer-theme attribute would leak at top level document

gotson:
I noticed that the UserSettings will apply the `data-viewer-theme` attributes on the top level document. When reusing the window after the reader was used, that attributes remains, but it should not.

this PR will make use of a `#root` element if exists instead, and else keep the existing behaviour for backward compatibility

### DIFF
--- a/src/model/user-settings/UserSettings.ts
+++ b/src/model/user-settings/UserSettings.ts
@@ -413,8 +413,10 @@ export class UserSettings implements IUserSettings {
         "html"
       ) as HTMLHtmlElement;
       if (html) {
-        const rootElement = document.documentElement;
-        const body = HTMLUtilities.findRequiredElement(rootElement, "body");
+        const rootElement =
+          HTMLUtilities.findElement(document, "#root") ||
+          document.documentElement;
+        const body = HTMLUtilities.findElement(html, "body");
 
         // // Apply publishers default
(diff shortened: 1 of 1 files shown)


aferditamuriqi:
@gotson i haven't tried this yet, but wouldn't this mean i have to have an element with id #root otherwise all of it will just be ignored?

aferditamuriqi:
> @gotson i haven't tried this yet, but wouldn't this mean i have to have an element with id #root otherwise all of it will just be ignored?

never mind i see what you are doing, will have it reviewed shortly 

i still question though when would there be an element with id root?

gotson:
> i still question though when would there be an element with id root?

it needs to be set by the application using the library. Setting the `root` element allows for controlled application of the theme onto other elements.

I do this in [Komga](https://github.com/gotson/komga/blob/898bc1981c665df8efe74b0cc483bea663d1941b/komga-webui/src/views/EpubReader.vue#L2-L2) by setting the top level component of that specific view as `root`, so the theme is applied on the whole view, not just the reader.

aferditamuriqi:
@gotson ok, i'll accept and merge this PR. since it doesn't break any backwards compatibility,

aferditamuriqi (approved):

PR #703, opened 2023-12-05
feat: add navigator api/events for chapterInfo and positionInfo updates

gotson:
I was trying to use my own component to display the bottom information of the reader:

> Page X of Y (Chapter Title)

I faced various issues while trying to use `api.updateCurrentLocation` to source the information i needed:
1. the function is called with a delay, which was clearly visible in the UI when rendering
2. the function is not called when changing the view mode from vertical scroll to paginated. If you were to navigate to a chapter via the TOC while in vertical scrolling, then changed the pagination to paginated, the api would not be called, and you would not have the correct information.

To solve 2. I tried to use `reader.currentLocator` just after changing the view mode, but since there are some internal timers inside the navigator, i would receive a locator with 0/0 for the page index/count.

This PR adds 2 new APIs / events, which are triggered at the same time/place than when the internal HTML components are updated. This allows to externalize the display with no drawbacks on delay.

### DIFF
--- a/src/navigator/IFrameNavigator.ts
+++ b/src/navigator/IFrameNavigator.ts
@@ -117,6 +117,8 @@ export interface NavigatorAPI {
   resourceAtEnd: any;
   resourceFitsScreen: any;
   updateCurrentLocation: any;
+  positionInfo: any;
+  chapterInfo: any;
   direction: any;
   onError?: (e: Error) => void;
 }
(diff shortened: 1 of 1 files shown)


aferditamuriqi:
@gotson i am reviewing and testing the issues you described here, i am not really understanding exactly the issue, especially about delay.
but one thing you are right, and it's not just when switching views, but changing any user settings there is no api.updateCurrentLocation called. i'll investigate why we did it this way but there was a reason for. i'll be back with more questions or comments soon.

gotson:
> i am not really understanding exactly the issue, especially about delay.

`updateCurrentLocation` is called from a debounce function called `savedPosition`, which has a 200ms delay

gotson:
@aferditamuriqi it seems this PR now has conflicts. Before i spent time fixing them, would you consider merging this ? Not sure if you had time to review that PR before.

aferditamuriqi:
Superseded by #1113 which implements positionInfo/chapterInfo callbacks with proper typing against the current codebase. Thank you for the contribution!

PR #705, opened 2023-12-07
feat: IFrameAttributes for fixed layout margin and shadow

gotson:
Currently the fixed layout applies a hardcoded margin of 100px on the side, as well as a drop shadow:
Screenshot 2023-12-07 at 16 51 38

This PR adds 2 new user settings:
- `fixedLayoutMargin` is a positive number that will control the margin. Defaults to the current value of 100.
- `fixedLayoutShadow` is a boolean to display or not the shadow. Defaults to the current value `true`.

Here is a screenshot with margins set at 0:
Screenshot 2023-12-07 at 17 48 57

Here is a screenshot with no shadow:
Screenshot 2023-12-07 at 17 49 19

### DIFF
--- a/src/navigator/IFrameNavigator.ts
+++ b/src/navigator/IFrameNavigator.ts
@@ -126,6 +126,8 @@ export interface IFrameAttributes {
   iframePaddingTop?: number;
   bottomInfoHeight?: number;
   sideNavPosition?: "left" | "right";
+  fixedLayoutMargin?: number;
+  fixedLayoutShadow?: boolean;
 }
 export interface IFrameNavigatorConfig {
   mainElement: HTMLElement;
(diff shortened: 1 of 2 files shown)


gotson:
I noticed today that on top of user settings, there is also `IFrameAttributes` and `reader.applyAttributes`. Given that function calls `handleResize`, which is where we apply the maring/shadow, it may be a better fit to use that instead of user settings.

Let me know what would work best, and i can amend the PR.

aferditamuriqi:
@gotson yes that would actually be better to use the IFrameAttributes instead of the user settings, as with user settings it's assumed that it will be injected into the html

gotson:
> @gotson yes that would actually be better to use the IFrameAttributes instead of the user settings, as with user settings it's assumed that it will be injected into the html

when i have time i will migrate this to IFrameAttributes

gotson:
@aferditamuriqi i have moved this to `IFrameAttributes`. This changes the default to no margin and no shadow, which i believe is a better default, but the attributes have been added in the dita example.

gotson:
@aferditamuriqi it seems this PR now has conflicts. Before i spent time fixing them, would you consider merging this ? Not sure if you had time to review that PR before.

aferditamuriqi:
Superseded by #1113 which implements fixedLayoutMargin/fixedLayoutShadow with backward-compatible defaults (100 and true). Thank you for the contribution!

PR #713, opened 2023-12-13
various iFrame fixes

gotson:
this PR fixes a few iFrame bugs i found while trying to make the fixed view double pages work.

The second iframe would be loaded twice in some cases, which did generate problem on Firefox only. When loading the book with a previously saved progress, the second iframe would not have any events registered. It would also not load when changing pages.

In the navigator, `handleIFrameLoad` would be registered for each iFrame. But within that function, the code would often either only care about `this.iframes[0]`, or loop through `this.iframes`. Since the function would be called twice when showing 2 iframes, duplicate effects would be generated, for instance all the event listeners would be duplicated.

I also disable the highlighter in fixed view mode. Whenever a click would occur on one of the page (usually images), the toolbox would be showing somehow. This fixes it.

### DIFF
--- a/src/modules/AnnotationModule.ts
+++ b/src/modules/AnnotationModule.ts
@@ -197,7 +197,7 @@ export class AnnotationModule implements ReaderModule {
     }, 200);
   }
 
-  initialize() {
+  initialize(iframe: HTMLIFrameElement) {
     return new Promise(async (resolve) => {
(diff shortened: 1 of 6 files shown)


aferditamuriqi:
@gotson i need some time to review and test this

aferditamuriqi (approved):

PR #714, opened 2023-12-13
feat: add support for reading direction

gotson:
This brings support for reading direction for fixed view. It is implemented as a User Setting that can be `auto`, `rtl` or `ltr`. That way it can be forced by the user to whatever direction. In `auto` mode the direction is read from the publication.

There is also a new api `direction` which can send `rtl` or `ltr`, and is used to know what is the effective reading direction applied. This is useful when using the `auto` mode, and can be used to adjust other elements of the page that are not owned/managed by R2D2BC.

I added support in the DITA sample reader.

### DIFF
--- a/src/model/user-settings/UserSettings.ts
+++ b/src/model/user-settings/UserSettings.ts
@@ -73,6 +73,7 @@ export interface IUserSettings {
   // publisherDefaults: boolean;
   textAlignment: number;
   columnCount: number;
+  direction: number;
   wordSpacing: number;
   letterSpacing: number;
   pageMargins: number;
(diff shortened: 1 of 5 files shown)


aferditamuriqi:
@gotson i am not so sure about the reading direction being settable, i think this needs some more work.

gotson:
> @gotson i am not so sure about the reading direction being settable, i think this needs some more work.

It's fairly possible to have publications without reading direction set, or incorrectly set.

src/model/user-settings/UserSettings.ts
```diff
@@ -361,6 +376,10 @@ export class UserSettings implements IUserSettings {
       "columnCount",
       ReadiumCSS.COLUMN_COUNT_KEY
     );
+    this.direction = await this.getPropertyAndFallback<Enumerable>(
+      "columnCount",
```
aferditamuriqi: this one seem incorrect though

gotson: indeed, i've corrected it

aferditamuriqi (approved):
LGTM

PR #758, opened 2024-01-31
Feature: add use memory store

aferditamuriqi:


gotson:
@aferditamuriqi  i feel like this is only changing the fallback store, but not the default store. I haven't tested, but looking at the code it seems it would still store settings in the local storage.

aferditamuriqi:
@gotson it doesn't I've tested it, just see my comment in the index file

gotson:
> @gotson it doesn't I've tested it, just see my comment in the index file

LGTM, tested and it works. Thanks.

aferditamuriqi:
> > @gotson it doesn't I've tested it, just see my comment in the index file
> 
> LGTM, tested and it works. Thanks.

you ok, with me using this PR instead of yours?

gotson:
> > > @gotson it doesn't I've tested it, just see my comment in the index file
> > 
> > 
> > LGTM, tested and it works. Thanks.
> 
> you ok, with me using this PR instead of yours?

sure

PR #784, opened 2024-02-17
fixes reading position in history on internal links

### LINKED ISSUE #676: History not kept when navigating with previous/next chapter buttons

aferditamuriqi:
Closes: #676
optimized this PR https://github.com/d-i-t-a/R2D2BC/pull/677


aferditamuriqi:
@gotson please review this PR and let me know if this solves your issue as well. the reason for this is, there is a scenario where if we just remove the check like in your PR, we will have double entries in history, for example, you select a chapter from TOC, and you don't scroll or page but click a link within that chapter, you will see 2 entries for progression 0 on that chapter. which would mean the user has to go back twice, in my PR i am accounting for that and also accounting for your fix, just in a different way

gotson:
@aferditamuriqi this seem to fix the original issue, i closed #677

PR #839, opened 2024-03-25
feat: add direct call to annotationModule `updateAnnotation`

### LINKED ISSUE #829: Add `updateAnnotation` to D2Reader instance properties
It would be helpful if there would be the possibility to call `annotationModule.updateAnnotation` directly from the `d2Reader` instance returned from `D2Reader.load()` method.

In my use case, I use a custom web component to show a modal where the user can type the note to add to an annotation, and such modal doesn't work like the SweetAlert library in the example of this repo, so I can't suspend the execution of `addAnnotation` or `addCommentToAnnotation` callback defined in the `annotations.api` config property.

Adding a public direct call to `annotationModule.updateAnnotation` (like it is for `addAnnotation` and `deleteAnnotation) would help me correctly handle the flow of my work.

MooDySmiles:
This pull request lets call the `updateAnnotation` method of `annotationModule` directly from the reader instance.

Closes #829

### DIFF
--- a/src/reader.ts
+++ b/src/reader.ts
@@ -522,6 +522,16 @@ export default class D2Reader {
   addAnnotation = async (highlight: Annotation) => {
     return (await this.annotationModule?.addAnnotation(highlight)) ?? false;
   };
+  /** 
+   * Update annotation
+   * 
+   * This should be used only when the add/delete of the annotation note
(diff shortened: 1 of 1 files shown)


aferditamuriqi (approved):
Thank you :)

PR #898, opened 2024-08-12
Fix for media overlay bugs in fixed layout books

jansc:
We experienced several problems when we were trying to play mediaoverlays in fixed layout books.

1. When the book is not opened on the first page, playing media overlays would only read every second page (only left or right spreadlink was set)
2. When reaching the last page of the book and the last page was in the left iframe, the mediaoverlay was also reading the previous page in the right iframe (missing initialization of currentSpreadLinks when switching pages)
3. When navigating to the first page after some mediaoverlay was played, the code would throw an unhandled exception (first page is always in right iframe, the left iframe might be about:blank, so `this.currentLinks[this.currentLinkIndex]?.Properties`  in the mediaoverlay module is not always available).

These commits seem to solve the above problems.

### DIFF
--- a/src/modules/mediaoverlays/MediaOverlayModule.ts
+++ b/src/modules/mediaoverlays/MediaOverlayModule.ts
@@ -203,7 +203,7 @@ export class MediaOverlayModule implements ReaderModule {
       this.settings.playing = true;
       if (
         this.audioElement &&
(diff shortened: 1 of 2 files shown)


aferditamuriqi (approved):
Looks good, thank you :)

PR #1010, opened 2024-11-05
Clear previous pages links

raulsi:
unset previous pages links.

if there is only 4 pages in book.
and after 4th page done it start playing the contents of 3rd page automatically, happens in 2 page layout as no 5th page exist.


aferditamuriqi:
Superseded by #1114 which applies the same fix against the current codebase. Thank you for the contribution!

